PR #2, opened 2018-05-16
[Code Commits] Notebook with a first proposal

jgbarah:


GeorgLink:
I removed myself as a reviewer, since I don't know what criteria to review this for.

sgoggins:
Working on bringing myself up to speed on a few different matters here, but I think for the good of our overall efforts: 

There are two things I am looking into : 
1. Making sure I understand the pattern laid out by this pull request.  Its the first of its kind, and will set a pattern we should all follow (or follow with tweaks maybe?)
2. I am thinking about how and to what extent we might want to explicitly or implicitly make the assessment of pull requests in this repository digestable by non-technologists.  

Practically, I also confess I don't know how to easily clone the notebook and run it inside of local instance of the repository systematically. On this count, I acknowledge feeling dumb.

jgbarah:
@GeorgLink said:
> I removed myself as a reviewer, since I don't know what criteria to review this for.

I think none of us have clear criteria yet, so any comment would be appreciated. But if you feel you're too much out of the loop to comment, don't worry ;-)

jgbarah:
@sgoggins said:
> Making sure I understand the pattern laid out by this pull request. Its the first of its kind, and will set a pattern we should all follow (or follow with tweaks maybe?)

This is the first intention to implement one of our metrics, as much precisely as possible, while showing at the same time the different choices that can be made (eg, counting merge commits or not).

> I am thinking about how and to what extent we might want to explicitly or implicitly make the assessment of pull requests in this repository digestable by non-technologists.

I think we could have like two kinds of prs: one more "general", for description of metrics, for example; and another one more specific, for their implementation. If we want to have detailed decriptions of the metrics, and reference implementations, I guess we need the latter. If we want "human readable" text for defining everything, I guess we need the former. This pr would be an example of the latter. 

> Practically, I also confess I don't know how to easily clone the notebook and run it inside of local instance of the repository systematically. 

Don't worry, I'm including some text in the notebook itself about how to run it. If you feel you need more guidance, just let me know.

PR #3, opened 2018-05-16
Cleaning up a bit, leaving only the files for GMD metrics.

jgbarah:
For now, only in the main directory, we still need to do this in subdirs.


germonprez:
Do you have the latest and greatest GMD from the Metrics repo?

Issue #5, opened 2018-05-17
GMD: issue resolution efficiency - What are abandoned issues?


aswanipranjal:
Following #1, What exactly are abandoned issues and how can they be calculated? 
- Are there any use cases available? 
- Is there a formula to calculate it?
- A solid definition?

Under: [Issue resolution efficiency](https://github.com/chaoss/metrics/blob/master/activity-metrics/issue-resolution-efficiency.md)

GeorgLink:
I'll throw this out there: 

I consider issues abandoned if there has been no activity within the last XX days. 

Activity can be new comments, new reactions, new pull requests, changed labels, changed milestones, or changed assignments. 

What the threshold of XX days is supposed to be, that can be a parameter chosen by the user, after all each community works differently, but I think 30 days might be a good start.

jgbarah:
I agree with @GeorgLink. However, maybe we should also have into account issues that are closed without fixing (eg, "won't fix"), but that's a bit more tricky (see below).

I think we can, for now, go with the approach @GeorgLink suggests, but in that case, I would change the name to "unattended issues", which is maybe more precise for that case. And is a useful metric, indeed.

We can also open a more general discussion about what exactly we want to track. I think this comes from an extension of the concept of "abandoned reviews" in Gerrit. There, it has a very precise meaning: reviews that were closed as "abandoned", meaning that the proposer of the patch is no longer working on it (thus, abandoning the patch, and signalling reviewers that no review is needed).

But for issues, that meaning is maybe difficult to capture, because it would be when somebody opens a ticket, and because whatever the reason, closes it before it is fixed. I think this is unusual, and I'm not sure to which extent useful... If any, could be a measure of engagement: I care so much, that I come to the issue tracking system to close this bug report, which was not addressed, but I found it was addresses somewhere else (for example).

So, in short, the "unattended" concept seems more appropriate to me.

jgbarah:
Re-reading the metric, maybe the ratio for efficiency would be better defined based on issues open / issues closed (instead of issues abandoned) during a certain period. In the end, it is related to the throughput, I think.

aswanipranjal:
@jgbarah: shouldn't it be: [number of issues closed (with a fix)] / [Number of total issues created]?

jgbarah:
> @jgbarah: shouldn't it be: [number of issues closed (with a fix)] / [Number of total issues created]?

There are two issues with that:

[...]

PR #12, opened 2018-06-12
Add detailed definition of Issue Resolution Efficiency

jgbarah:
After the discussion in #5, this tries to apply the results and consensus of that discussion, to start defining this metric.

Closes: #5.

### DIFF
--- a/activity-metrics/issue-resolution-efficiency.md
+++ b/activity-metrics/issue-resolution-efficiency.md
@@ -1,25 +1,62 @@
 # Issue Resolution Efficiency
 
 ## 1. Description
-What is the number of closed issues/number of abandoned issues?
+What is the number of closed issues/number of open issues?
 
 ## 2. Use Cases
-Provide examples of how the metric might inform different stakeholders through use cases.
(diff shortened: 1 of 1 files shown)


jgbarah:
Once we agree on this, we can explore a similar metric for code reviews.

activity-metrics/issue-resolution-efficiency.md
```diff
@@ -1,25 +1,62 @@
...
 
 ## 3. Formula
-A generic formula (in pseudo code) to generate the metric.
+The definition is for a time period (for example, for a certain month):
+
+**Definitions:**
+
+* `issues_closed`: Total number of issues closed during the period
+* `issues_opened`: Total number of issues opened during the period
+* `issues_backlog`: Total number of issues open at the beginning of the period (backlog of open issues)
```
sgoggins: Does it make sense to have this be the beginning or the end? I think either one would generate the same number because you are going to simply be adding "open now, but not opened during this period" in effect. I am tryin to think of when the number would be different at the beginning or end if its a windowed time period?

PR #21, opened 2018-09-21
add two use cases to forks

GeorgLink:
Hi GMD workgroup,

I just had an interview and we discussed forks.
I am sure there is literature to back up the two use cases, but I wanted to make sure to capture this thought first and we can improve on it later.

Best,
Georg


sgoggins:
We are no longer using this repository. Can you issue the pull request against the "wg-gmd" branch of the metrics repo?

GeorgLink:
Will do.

An issue template would be great, that informs someone trying to open an issue or pull request that this is not the place.

jgbarah:
I'm reopening this pr, since I think we agreed to follow up work here.

sgoggins:
looks good to me. Just fix the directory.

jgbarah:
@GeorgLink maybe you can amend the commit to avoid the current conflict? I guess it is a matter of changing the file name to the new one for this metric.

GeorgLink:
I guess so. I'll have a look next week.

GeorgLink:
I updated the branch to include the latest master version... 

When I `git diff` compare locally, it shows correctly that I only made a minor change.... this pull request shows way more.

sgoggins:
Looks like this is a simple change, but the commits look more significant because things moved.

Issue #7, opened 2018-05-17
GMD: Closed Issue Resolution Duration


aswanipranjal:
Following #1,
[Closed issue resolution duration](https://github.com/chaoss/metrics/blob/master/activity-metrics/closed-issue-resolution-duration.md). As the name suggests: is this the time since an issue has been resolved? Again, do we look at all the issues or do we take an average?

Also, the definitions might need some improvement.

jgbarah:
I agree with @GeorgLink for the definition. To make it more clear from the name, I would change the name to "time-to-resolution of closed issues", or maybe more clear, "time-to-close of closed issues".

WRT aggregation, same comments that those in #6 apply.

In addition, calculating it for specific periods of time (as @GeorgLink) suggests, is sensible, but maybe that depends on the granularity of the "evolution periods" that we're using in an specific case, something that I guess applies to almost any metric.

jgbarah:
I think this is closed now, by #24.

Issue #4, opened 2018-05-16
Clean-up repository


jgbarah:
We still have some Markdown files in the activity-metrics directory corresponding to metrics that are not used in GMD. I think we should remove them. I didn't just in case I'm missing something... Please, anyone interested, comment here. If we agree, I can do the job later (but prs by anybody are welcome).

klumb:
As part of aligning and defining CHAOSS processes, I can start working on this.

jgbarah:
I think this is done for now, thanks to the work by @klumb and others. Thanks!!!

PR #24, opened 2018-10-06
Repository Structure and File Name Convention

klumb:
There is a lot here and I took some liberties on naming the metrics files so it may be helpful to discuss the 30 metrics we have identified so far (according to the goal documents) in the next meeting to make certain we have shared understanding of what they are and the names are accurate. 

(or comments/edits here)

For the most part, I left the metrics files untouched other than changing the name or creating a file if it was missing.

I moved the metrics that aren't linked in the goal pages to an archived metrics folder.

I created a focus area folder to hold the goals and use cases. I believe this is the structure that was discussed.

I also made minor changes to the repository structure and some of the organization documents. 

Resolves:
#24 
#7 
#4 

@sgoggins 
@jgbarah 
Please Review


klumb:
I am not sure what .DS_Store was it showed up 5 or 6 months ago. I deleted it.

klumb:
Sorry, slipped and hit wrong button

jgbarah:
@klumb, anything about this?:

> Why are we using numbers for starting folder names? (I mean, such as 1_template_folder instead of template_folder

klumb:
@jgbarah, also not needed.

PR #26, opened 2018-10-17
[use_cases] Add directory and description for use cases.

jgbarah:
This is a preliminar version, to start working with use cases.

### DIFF
--- a/use_cases/README.md
+++ b/use_cases/README.md
@@ -1,0 +1,72 @@
+# GMD Working group use cases
+
+This directory is for describing use cases related
+to the use of GMD (growth-maturity-decline) metrics.
+The main aim of these use cases is to provide information on
+how people are using, or intending to use, GMD metrics.
+
+The CHAOSS GMD working group has identified this kind of use cases as one of its key
+goals, so that we can define metrics that are useful in the real world.
+
+## What is a use case in this context?
+
+A use case should show, in some specific aspect,
+one or more high level goals, that can be refined in questions,
+which could be answered with metrics.
+
+If possible, the use case should identify some of the metrics that could lead
+to those answers, but that is not required in the first version of the use case
+(the CHAOSS community could later help to identify those metrics)
+
+## Who can propose a use case?
+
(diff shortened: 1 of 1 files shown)


sgoggins:
Isn't this what we would put into the "Focus Areas" Folders if we are following the D&I Structure?

sgoggins:
Following up on my comment just now, Underneath "Focus Areas" I see the specific metrics defined. I know we have discussed using Use Cases, but I do not see them in the D&I Structure. @GeorgLink : Can you advise regarding how D&I is handling use case definitions and if any are committed to the repository yet? Or, if the use cases are embedded in your metrics definitions?

jgbarah:
As far as I know, D&I are not using use cases in this way. The idea here came from several people in CHAOSS having interesting experiences on using (or willing to use) metrics, which could be useful for defining the questions and framing the metrics.

The idea of these use case is to be orthogonal to the "focus area | goal | question | metric" structure (therefore the different directory).

GeorgLink:
In D&I, we do not use the term 'use case'. What we include on our resource pages are 'objectives', which are not full fledged use cases but provide rationale for why to use metrics. [Example][1]

How does the use case compare to our idea of the blog posts series [Metrics in Use][2] (which we still need to launch)?

[1]: https://github.com/chaoss/wg-diversity-inclusion/blob/master/focus_areas/project_and_community/documentation.md
[2]: https://docs.google.com/document/d/1p9FZM6rixjiEsxXQ7Ij-mbGCJKm_OrOQ6nd3oIBRnto/edit

jgbarah:
> How does the use case compare to our idea of the blog posts series Metrics in Use (which we still need to launch)?

I think most use cases could be the basis for a blog post.

jgbarah:
> One suggestion: Point to our blog post series as a possibility for disseminating the use case. https://docs.google.com/document/d/1p9FZM6rixjiEsxXQ7Ij-mbGCJKm_OrOQ6nd3oIBRnto/edit#heading=h.i08ikslakwjv

Good idea. I'm going to do it.

PR #27, opened 2018-10-18
[meetings] Agenda and minutes for Oct 10th meeting

jgbarah:


klumb:
We could remove the reference links at the bottom. Looks good otherwise.

jgbarah:
> We could remove the reference links at the bottom. Looks good otherwise.

I thought they could be interesting as a reference, but if you prefer, I will delete them.

GeorgLink:
Why remove the reference links?

I think they are good for building a shared language and understanding.

klumb:
The reference links aren't specific to the meeting. The text comes from the email I sent with the notes in it. They are part of a template I am creating for the recurring agenda and notes emails so all future versions of my notes/agendas will also include it. 

Question - Are we appending notes to this document or are we replacing them? If we are replacing them then I see no issue in leaving them. If we are appending multiple meeting dates here, then the links will be redundant.

GeorgLink:
Maybe I missed something.

Will notes be for the entire document or only live within a section of a
specific meeting?

I favor the latter and do not mind repetitions if it i accurately reflects
what is on the mind of members or was mentioned during the meeting.

klumb:
I think we should fast track these notes pull requests

PR #29, opened 2018-10-25
[use_cases] Be more specific about GitHub issues and alternatives

jgbarah:
After the discussion during the last GMD meeting, I'm adding some text on using Google Docs when appropriate.

I also added some more details and links to documentation for GitHub issues and pull requests.

### DIFF
--- a/use_cases/README.md
+++ b/use_cases/README.md
@@ -26,7 +26,7 @@ There is no requirement of being a regular contributor of the working group.
 
 ## How can I propose a use case?
 
-Open an issue in the [repository for this GMD working group](https://github.com/chaoss/wg-gmd),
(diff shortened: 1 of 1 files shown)


use_cases/README.md
```diff
@@ -35,14 +35,21 @@ In the description of the issue, write one or more high level goals of the use c
 and if possible a list of the questions that could be relevant for
 reaching that goal.
```
sgoggins: I think we should require question mapping, otherwise I think we will struggle to connect the use cases to metrics.

jgbarah: Do you mean removing "if possible" in that sentence?

jgbarah: I hope that's what you meant, @sgoggins ;-) I amended the text with a new commit, and I'm merging all of this. We can always amend it later, if needed, but this way we make this visible to everyone.

PR #30, opened 2018-10-25
[use_cases] Template for use cases

jgbarah:
This is a proposal for a template for use cases, after our discussion on this matter during the last GMD meeting.

I've used the SPDX use cases as a template, adapting it to our case.

### DIFF
--- a/use_cases/template.md
+++ b/use_cases/template.md
@@ -68,4 +68,4 @@ description of the metric, and which questions it helps to answer.
 
 ## Comments
 
-Any other relevant comment.
+Any other relevant comment. For example, specific threasholds for metrics could be provided, such as "if XXX metric reaches YYY value, the benefits for the project are clear".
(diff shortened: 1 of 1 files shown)


GeorgLink:
Thanks for putting this together Jesus. The template looks good.

What about a section that discusses the value of a metric for decision making? E.g. if value is above 80th percentile, then this indicates an amazing investment opportunity.

jgbarah:
Thanks, @GeorgLink.

For the value of the metric, maybe we could use for now the comments section? For now, I was intending to focus on the description of the goals and questions, so that anyone can write their use cases, even if the use of specific metrics is not completely clear. I'm going to include it as a possible comment for now, while I wait for your feedback.

jgbarah:
Thanks for your comments, @GeorgLink @sgoggins 
Merging...

PR #34, opened 2018-10-30
Rename metric detail files

ccarterlandis:
Some metrics had confusing names. For example, the metric `lines-of-code-changed` was defined in a file called `pull-requests-lines-of-code-changed.md`, despite the fact the number lines of code changed is not directly related to pull requests. I have updated these metrics to be reflect such inconsistencies.

In addition, other files were slightly renamed duplicates - in the interest of keeping things tidy, I went ahead and deleted the duplicates. 

I also tried to standardize the format of metric names as well as make them easier to understand without reading the detail file.


sgoggins:
This is some house keeping. Everything code related had a "pull requests" preface, which made no sense in a number of cases. Also got rid of some metrics that obviously duplicated each other.

jgbarah:
I have nothing against the changes, but I don't understand its rationale. Some of the files I don't find them in the goals files. For example, I cannot find `pull-requests-review.md` in the [Code Development goal](https://github.com/chaoss/wg-gmd/blob/master/focus_areas/code_development/goal_code_development.md). Maybe I'm missing something?

ccarterlandis:
I saw some of those too, but I wasn't sure if they were a metrics that still needed a home, or if someone else was perhaps working on them, so I elected to leave them alone and tried to mostly clean up cruft and rename metrics that I thought could benefit from a renaming. Sorry for the confusion @jgbarah! Please let me know if there's anything else!

PR #36, opened 2018-10-31
[README] Proposal for a more informative README.md

jgbarah:
I've tried to put together all the issues that we commented during our last meeting, on how to contribute, when we meet, how we're structured, the usual process to propose changes, etc. I propose this just as a first step, further pull requests could refine everything in this file.

For some file/directory names I'm proposing another pull request, simplifying their names (I'm already using the "new" names here).

PR #37, opened 2018-10-31
Rename some directories, to simplify structure

jgbarah:
The main reason is to use shorter names. The changes are:

* example_implementations -> implementations
* detail_metrics_methods -> metrics

This is already referenced in changes in #36, if possible please merge before it (but this is really independent from it).


klumb:
Long folder and file names are one of my foibles.  This change is good, I vote to merge.

PR #38, opened 2018-10-31
Refining meeting notes

jgbarah:
* Removed some information that is redundant with README.md.
* Restructured some headers, to use fewer levels (when possible).
* Template for future meeting notes, at the end of the file.


klumb:
This is good, I vote to merge it.

PR #40, opened 2018-10-31
[use_cases] Add section on vocabulary

jgbarah:
In some cases we need to define specific terms used in the use case. We add a new vocabulary section for that.


rpaik:
For "core contributors", maybe we should follow the example we have at https://github.com/chaoss/grimoirelab-sigils/blob/master/doc/community-structure.md?  I think we need to add a timeline/time period of contribution.  Something like 6 months or 1 year makes sense to me....

PR #42, opened 2018-10-31
meeting notes october 31 2018

klumb:
These are the meeting notes for October 31st - Feel free to edit or add content if I missed something.


jgbarah:
Merging,since there is no comment against and I already approved.

Issue #33, opened 2018-10-30
Use Case: Efficiency in reviewing contributions from first-time contributors


rpaik:
A first-time contributor could be easily discouraged if no one reviews/acknowledges their work in a timely manner. Following are number of questions that can help evaluate responsiveness for first-time contributors. 

- What percentage of first-time contributions are accepted?
- What percentage of contributions from first-time contributors are accepted vs. `core developers`?
- How long does it take (e.g. average or median) for first-time contributions to be `resolved` (e.g. accepted or declined)?
- How long does it take (e.g. average or median) for contributions from `core developers` to be `resolved` (e.g. accepted or declined)?
- How long does it take (e.g. average or media) for first-time contributions to be responded to (e.g. acknowledgment, first review, etc.)? 
- How long does it take (e.g. average or median) for contributions from `core developers` to be responded to (e.g. acknowledgment, first review, etc.)? 
- Does the acceptance of first-time contribution have an effect on future contributions from the same contributor? 
- Does the responsiveness to first-time contribution (e.g. average/median time for acknowledgement, first review, etc.) have an effect on future contributions from the same contributor?

GeorgLink:
Thanks @rpaik 

First-time contributors seem to be easily defined as those who have no prior contributions.

Do we have a definition for `core developers` (or `core contributors`)?

jgbarah:
@GeorgLink said:

> Do we have a definition for core developers (or core contributors)?

Good question. Maybe we could work with an intuitive idea of what core developers (contributors) are, as those who perform most of the contributions, until we define the specific metrics.

In any case, we could define that either:

* In administrative terms: developers with admin privileges, or nominated as such by the project (which is difficult to generalize, but in some specific cases, such as GitHub, could be particularized as people with admin and/or merge rights on the repository)
* In quantitative terrms. Following the 80-20 rule, we could consider people producing 80% of the contributions over a certain period of time (eg, since the start of the project, or during the last six months).

But maybe we can think about the questions, and then metrics, based on that we can define "core", to sidstep that issue...

In fact, I wonder if that could lead to another use case: "Knowledge about who are core contributors to a project". What do you think?

jgbarah:
When particularizing, we should also define "contribution". For example, a commit is (I think) clearly a contribution. Maybe, at least for a start, we could concentrate on this. But there are other kinds of conributions... What do you think @rpaik: for focusing, we consider only code contributions in the form of commits, or we try to extend to other kinds of contributions?

[...]

PR #47, opened 2018-11-13
[use_cases] First draft of "code contributions by new contributors"

jgbarah:
After some discussion in issue #33, this is the first draft, based on the proposal by Ray, for formalizing this use case.

Please, consider if it is "good enough" for a first take, we can refine it later in other pull requests.

It does not include metrics yet, I think we should first decide if questions are good enough.


germonprez:
@sgoggins, can you review this?

PR #63, opened 2018-12-19
[use_cases] First try at adding metrics to a use case

jgbarah:
This is a first try at adding metrics to a use case (#33), and the problems that we find.

In this case, the proposed metric is still not in our collection, therefore the TODO instead of the link,
(maybe better a link to `../metrics/still_undefined.md`?).

Then, the list of questions in which this metric is relevant, and how it would be used to answer the question.

Any feedback is appreciated.


jgbarah:
@sgoggins @rpaik (or any other one): any comment?

Issue #60, opened 2018-12-12
GMD Goals for 2019


klumb:
We are creating a gmd goals document.

Here is the working document
https://docs.google.com/document/d/1CkzlCT1hi9OI1SQf_GAZSsNv9o3Po25x8wC5G424d2U/edit

jgbarah:
I share the general lines I see in that document. Should we convert it into a PR, and eventually merge it? I suggest naming it ROADMAP.md (which is customary in some repos), and then making it evolve to a real roadmap if we want.

jgbarah:
We already have a first roadmap, so we can close this.

PR #74, opened 2019-01-23
First version of a roadmap

jgbarah:
First version of a roadmap, based on the document we prepared in #60.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,87 @@
+# Roadmap: CHAOSS GMD WG Goals for 2019
+
+## Problem Statement: 
+
+A wealth of data can be obtained from software development repositories,
+which could be useful to better understand the projects running them.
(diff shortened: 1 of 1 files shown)


ROADMAP.md
```diff
@@ -0,0 +1,87 @@
...
+### Goal #4: Have more use cases
+
+We need use cases to inform how people are using, or intending to use,
+metrics for understanding their projects.
+This should be one of the sources of information to write our
+goal-question-metric analysis in the focus areas documents.
+
+* We are aiming to have more use cases fully described.
+
+### Goal #5:  Grow the work group
```
GeorgLink: Goal #5: Maybe GMD could also do some introspection and document workflows to make them accessible to new members. Maybe that's part of the 'onboarding documents', but I think of it more as governance documents that guide every-day work.

PR #79, opened 2019-01-29
[metrics] Minor fix to link to implementation for Code_Changes

jgbarah:
The link was wrong. In addition, I'm adding a direct link to the Binder version, which is more user friendly.


jgbarah:
Thx! Merging...

PR #86, opened 2019-02-15
Update issue_resolution.md

germonprez:
Again, just aligning names between what's in the table and what is on the detailed pages

### DIFF
--- a/focus_areas/issue_resolution.md
+++ b/focus_areas/issue_resolution.md
@@ -7,7 +7,7 @@ Metric | Question
 [Open Issues](../metrics/issues-open.md) | What is the number of open issues?
 [Closed Issues](../metrics/issues-closed.md) | What is the number of closed issues?
 [Open Issue Age](../metrics/issues-open-age.md) | What is the the age of open issues?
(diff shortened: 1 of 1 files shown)


focus_areas/issue_resolution.md
```diff
@@ -7,7 +7,7 @@ Metric | Question
 [Open Issues](../metrics/issues-open.md) | What is the number of open issues?
 [Closed Issues](../metrics/issues-closed.md) | What is the number of closed issues?
 [Open Issue Age](../metrics/issues-open-age.md) | What is the the age of open issues?
-[First Response to Issue Duration](../metrics/issues-maintainer-response-duration.md) | What is the duration of time for a first response to an issue?
+[Issue Response Rate](../metrics/issues-maintainer-response-duration.md) | What is the duration of time for a first response to an issue?
```
jgbarah: Rate is not more like the speed at which something changes? In this case, the metric is not about how this changed with respect to the previous period, or something like that, but how long does it take to first respond... I agree the current name is confusing, but this could carry its own confusion. What about "Issue Response Time"?

sgoggins: I agree on this .  ... "Issue Response Time".   I asked Matt to update the pull request accordingly.

PR #88, opened 2019-02-16
pass sort=True to df.append() to avoid warning

Polaris000:
While running the script, I noticed that the sort warning popped up for the append() method. [Link to pandas documentation mentioning the change](http://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.append.html?highlight=append#pandas.DataFrame.append). 
screenshot from 2019-02-16 21-21-36

### DIFF
--- a/implementations/Code_Changes-Git.ipynb
+++ b/implementations/Code_Changes-Git.ipynb
@@ -130,7 +130,7 @@
     "                commit = json.loads(line)\n",
     "                commits.append(self._summary(repo=commit['origin'],\n",
     "                                             cdata=commit['data']))\n",
-    "        self.df = self.df.append(commits)\n",
+    "        self.df = self.df.append(commits, sort=True)\n",
     "        self.df['author_date'] = pd.to_datetime(self.df['author_date'], utc=True)\n",
     "        self.df['commit_date'] = pd.to_datetime(self.df['commit_date'], utc=True)\n",
     "        \n",


Polaris000:
@jgbarah please have a look

jgbarah:
Thanks! It would be worth exploring if `sort=False` (which is less time-consuming) is also ok (I think it is, but I'm not completely sure). I would appreciate if you can have a look and find out.

But for now, I'm merging this, since it is better than the current situation ;-)

Polaris000:
Thanks for the reply. I'll find out

Polaris000:
So first I tested the append method directly by timing the creation of a Code_Changes object for different counts as shown below:

```python
sort_true_times = list()
sort_false_times = list()

def time_test(sort):
    changes = Code_Changes('git-commits.json', sort)

for i in range(100):
    sort_true_times.append(timeit.timeit("time_test(True)", setup="from __main__ import time_test", number=i))
    sort_false_times.append(timeit.timeit("time_test(False)", setup="from __main__ import time_test", number=i))
    print(i, end=" >> ")
    
plt.plot(sort_true_times, 'r')
plt.plot(sort_false_times, 'b')
plt.show()
```


**red -> sort=True
blue -> sort=False**

The output:
image

Next, I checked for each class method, which were making use of self.df:
```python
sort_true_times = list()
sort_false_times = list()

changes_true = Code_Changes('git-commits.json', True)
changes_false = Code_Changes('git-commits.json', False)

functions1 = [
                "changes_true.total_count()", 
                "changes_true.count(since='2018-01-01', until='2018-07-01')",
                "changes_true.by_month()"
             ]

functions2 = [ 
                "changes_false.total_count()", 
                "changes_false.count(since='2018-01-01', until='2018-07-01')",
                "changes_false.by_month()"
             ]

for i in range(3):
    sort_true_times.append(timeit.timeit(functions1[i], globals=globals(), number=500))
    sort_false_times.append(timeit.timeit(functions2[i], globals=globals(), number=500))
    print(i, end=" >> ")
    
plt.plot(sort_true_times, 'r')
plt.plot(sort_false_times, 'b')
plt.show()
```

The output:
0 -> total_count()
1 -> count(since='2018-01-01', until='2018-07-01')
2 -> by_month()


**red -> sort=True
blue -> sort=False**


image

jgbarah:
Thanks for the detailed analysis!

Issue #48, opened 2018-11-14
Use case: Characterize a participant, both within a given project and across projects.


kfogel:
This may be a long-ish description -- I'm porting a formerly private conversation with @jgbarah over to here.  (If there's some more appropriate forum, please let me know!)

To set some context: almost all the information I'd want to find out about a project is actually built up from the activity of the *people* in that project, so I've come to think that most of the metrics I'm looking for are about people primarily, and projects secondarily.

* **Use Case:** Evaluate whether a person should be offered commit access or whatever the equivalent is for the kind of work they're doing (e.g., maybe it's actually "a vote on the steering committee" or "an editorship position in the documentation" or whatever).  The point is, existing maintainers often find themselves evaluating a person's activity in order to determine whether that person should be invited into some kind of formal maintainer role.

  The metrics required to do this thoroughly are interesting:

  - The data collection system needs a concept of the person's unified identity.  I.e., this account over here in the repository (which may or may not use the same authn system as the issue tracker) is the same as that account over there in the wiki... is the same person as these two email addresses on the mailing list... all of which are the same person as so-and-so over in the chat platform, etc, etc.  (Obviously it will not always be possible to determine these correspondences automatically, so there needs to be a UI where some knowledgeable human can come in and just *tell* the system that two accounts both belong to the same person, after which the system can faithfully remember that fact.)
  - A time-based view of when the person is active.  I almost care *that* they are active more than *how* they're active.  It's hard to tell, in an automated way, whether a given commit was more of a contribution than an answer given in Zulip or IRC, for example.  But just looking at *when* and *how often* the person did stuff tells you a lot.
  - When the person is responding to someone else in a thread (whether it be in the issue tracker, mailing list, or somewhere else), what is the range of latencies of the person's responses?

So that's what I want: an ability to look at a list of participants in a project, and see what they've been up to.  Looking at a person's activity within a single project would already be useful.  Ideally, I'd like to look at that person's activity in as many projects as possible.  For example:

* **Use Case:** If I knew that a particular bug reporter had a tendency to report bugs that resulted in associated bug fixes that were successfully merged, with the ticket closed as "FIXED" (or whatever the equivalent state is in that tracker), then when that person reported a bug in my project, I might pay attention more quickly.

[...]

PR #91, opened 2019-02-20
[use_cases] New use case: Hints for maintainership position

jgbarah:
This is the first iteration of this use case. It does not include metrics yet, until we agree on the questions (as it is now, only a single question).

Any comment is appreciated, but comments by @kfogel are specially appreciated, since he started #48, which is the issue where we are discussing this use case.

### DIFF
--- a/use_cases/maintenership-position.md
+++ b/use_cases/maintenership-position.md
@@ -1,0 +1,75 @@
+# Hints to help decide suitability for maintenership position
+
+## Description
+
+In some cases, projects need to evaluate and decide whether a person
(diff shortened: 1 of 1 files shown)


jgbarah:
@kfogel any feedback?

jgbarah:
Since we're having no feedback, I'm proposing to merge this anyway, and if needed, work later on the resulting files. Do you agree?

foundjem:
I agree.

GeorgLink:
Merged as a first version of the use case. continue discussion in $48

PR #97, opened 2019-03-05
[metrics] Add metrics related to the question Proposals.

jgbarah:
Rename files corresponding to legacy metrics that sort of correspond to the metrics considered in this change.

Fill the data for the following metrics for code development:

* Proposals
* Proposals_Accepted
* Proposals_Declined


metrics/Proposals.md
```diff
@@ -0,0 +1,124 @@
...
+as long as it proposes changes to source code files.
+
+The date of the proposal can be defined (for considering it in a period or not)
+as the date in which the merge request was submitted.
+
+### GitLab parameters
+
+None.
+
+## Specific description: Gerrit
```
sgoggins: There is an existing specific description for GHTorrent I think we should integrate. 

###  GHTorrent

```SQL
SELECT count(distinct pull_request_id) as num_still_open, projects.name as project_name, projects.url as url
FROM msr14.pull_request_history
    join pull_requests on pull_request_history.pull_request_id = pull_requests.id
    join projects on pull_requests.base_repo_id = projects.id
where pull_request_id not in
    (SELECT pull_request_id
    FROM msr14.pull_request_history
    where action = 'closed')
group by projects.id
```

metrics/Proposals.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Known Implementations
+
+* [Grimoirelab](https://chaoss.github.io/grimoirelab). Enriched index for
+GitHub Pull Requests, GitLab Merge Requests,
+and Gerrit repositories is composed of one item per proposal,
+which makes it basically correspond to this metric when counted.
+The GitHub Pull Requests, GitLab Merge Requests, and Gerrit summary panels,
+available out of the box, provide exactly that.
+
```
sgoggins: There is also an existing implementation for these in Augur: 
[Augur](http://dev.augurlabs.io/metrics_status)

metrics/Proposals_Accepted.md
```diff
@@ -0,0 +1,151 @@
...
+
+* Volume of coding activity.
+
+    Accepted code proposals are a proxy for the activity in a project.
+    By counting accepted code proposals in the set of repositories corresponding
+    to a project, you can have an idea of the overall coding activity in
+    that project that leads to actual changes.
+    Of course, this metric is not the only one that should be
+    used to track volume of coding activity.
+
```
sgoggins: I think the community manager use case is relevant for these metrics.

metrics/Proposals_Accepted.md
```diff
@@ -0,0 +1,151 @@
...
+These could be represented as bar charts, with time running in the X axis.
+Each bar would represent accepted proposals to change the code
+during a certain period (eg, a month).
+
+## Reference Implementation
+
+[ To be done. ]
+
+## Known Implementations
+
```
sgoggins: - [Augur](http://dev.augurlabs.io/metrics_status)

metrics/Proposals_Declined.md
```diff
@@ -0,0 +1,127 @@
...
+These could be represented as bar charts, with time running in the X axis.
+Each bar would represent declined proposals during a certain period
+(eg, a month).
+
+## Reference Implementation
+
+[ To be done. ]
+
+## Known Implementations
+
```
sgoggins: - [Augur](http://dev.augurlabs.io/metrics_status)

metrics/pull-requests-closed.md
```diff
@@ -1,18 +0,0 @@
-# Pull Requests Closed
```
sgoggins: I am concerned about deleting metric names that we have been working with. 
I agree with this change in name because it is more general, but we only just aligned these old names with the main metrics repository; So I think there is a cascading effect ... a contract of sorts, between us and the main metrics repo that has to abided along with this deletion. 
1. Rename metric here
2. Implement markdown of renamed metric new names
3. implement a pull request into the main metrics repository simultaneously to ensure ongoing alignment. 

If people know what a thing is called and have referenced it I think we cause ourselves problems by just taking it away.

jgbarah: We haven't released yet, so anything up to now is subject to be changed, i think... That's the reason why I don't understand a lot the metrics repo. But I'm happy to follow this process you say if you feel it helps. I, or somebody else, could do 2. and 3. on the metrics repo once the change is accepted here.

metrics/Proposals_Accepted.md
jgbarah: That's fine with me. Would you mind submiting a pull request with that, once this gets approved? Just to move on...

metrics/Proposals.md
jgbarah: Reading the SQL, this seems like a different metric. the filter is for `pull_request_id not in ... where action = 'closed'`, which I thin is therefore counting pull requests that are still not closed. The metric defined here is for pull requests initiated during a certain period. Besides, I don't a filter on time.

This said, I have two more general concerns:

* GHTorrent is not exactly a data source. GitHub is, GHTorrent is providing secondary information on what the original data source provides. From this point of view, the GHTorrent database is more like the GrimoireLab database: it retrieves data from the original data source, and organizes it in some specific way.
* This part of the specification was not (at least in my mind) to provide reference implementations. For that we have an specific section, below. This was for discussing specific peculiarities of the data source, and in this case, that is done for the three data sources that we're proposing for change proposals: GitHub pull requests, GitLab merge requests, and Gerrit changesets.

Maybe we could discuss, from a more general point of view, if we want to change the format for the description of the metric....

metrics/Proposals.md
jgbarah: I see two entries for a metric "Pull_Requests_Open", but both seem to link to the [same definition](https://github.com/augurlabs/metrics/blob/wg-gmd/activity-metrics/pull-requests-open.md), which corresponds to the SQL in the comment above. As I understand, this corresponds to pull requests open as of now (well, as of the last update in the GHTorrent database), not to this metric.

Maybe I'm missing another metric which corresponds to this one?

metrics/Proposals_Accepted.md
jgbarah: Which one it is? (I'm sorry, I couldn't find it, very likely I just didn't identify it by name).

metrics/Proposals_Declined.md
jgbarah: Same... I'm sorry, but I couldn't find it.

jgbarah:
Thanks a lot, @sgoggins. Please have a look at my comments.

metrics/Proposals_Accepted.md
sgoggins: Yes.

metrics/Proposals.md
jgbarah: I see your point too. Let's discuss in a moment, during our meeting...

PR #108, opened 2019-03-12
[metrics] Add details for metrics for the question Issues.

jgbarah:
Metrics related to the goal Activity, question Issues, in the code development focus area.

First version, ready to release from my side.


metrics/issues-open.md
```diff
@@ -1,54 +1,55 @@
-# Open Issues
```
sgoggins: I think the text in here is applicable to the Issues-Active metric. We shouldn't lose this work.

focus_areas/code_development.md
```diff
@@ -43,11 +43,14 @@ are happening during a certain time period?
 * Question **Issues**: How many issues related to the source code
 are happening during a certain time period?
 
-  * Metric **Issues**(Period): Number of new issues related to the source code.
+  * Metric **Issues**(Period): Number of new issues related to the source code
```
sgoggins: I think a better name might be Issues-Opened.

metrics/issues-open.md
jgbarah: We have Open Issues somewhere else, I think...

Active issues are issues active (having some activity, such as comments, changes in state) during a certain period.

Open issues are issues open at a certain point in time.

So, both metrics are different.... But as I said, I guess we still have the text for open issues elsewhere.

focus_areas/code_development.md
jgbarah: We can change it. I just tried to be in sync with other definitions such as "Proposals" or "Commits", and being minimalist with the naming. But I'm happy to change if you feel that's better. Any other opinions?

metrics/issues-open.md
jgbarah: OK, just clarify in this respect, since we have three different merics that could be mistaken (of of them mentioned in the code development focus area):

* Issues: Number of new issues (opened during a certain period)
* Issues_Active: Number of issues that showed some activity (during a certain period)
* Issue_Backlog: How many issues are still open? (at a given point in time)

This patch deals with the first two. In particular, Issues, as you mention below, could also be Issues_Opened, or also, New_Issues. The third one is "issues that remain open right now (or at any other given point in time), and from that point of view, could also be named "Issues_Open", but that could be confusing with "Issues_Opened". This is the reason why I was trying to avoid "open" and "opened" in the metric name.

jgbarah: And in any case, @sgoggins wrt to your original message, i agree: `issues-open.md` in this change is still in the repository, and I plan to reuse it for "Issues_Backlog" (see the file for the code development focus area).

focus_areas/code_development.md
jgbarah: After thinking about it, and because of the reasons mentioned in the previous comments, I agree with finding a more specific name. But to avoid the problems of "open" vs "opened", I'm changing this to "Issues_New", which I hope captures the idea of "issues opened (new) during a certain period". Do you think this is good enough?

metrics/issues-open.md
rsanchez-seerene: New issues normally are associated to a time frame namely in the last x hours, days or that have not been acknowledged (In a monitoring system). Thus I would propose to pick a time frame, although there will be the caveat that some issues may be resolved within the chosen time frame and technically would no longer be new issues.
--------------------
Metrics 2 and 3 could be misinterpreted since Active and Backlog are interchangeable terms in an operational environment but in an Agile environment "represents an option the team has for delivering a specific outcome rather than a commitment".
 
Pending, unresolved, fixed and closed are more finite status indicators that I think most understand.
--------------------
Two important metrics that I personally find of value are Issue_type and Issue_severity which help prioritization and picture the response to the respective categories or combinations.

jgbarah:
Merging, since nobody opposed.

PR #118, opened 2019-03-27
Removing value and risk as focus areas for this wg

jgbarah:
Now that we have wg-risk and wg-value, it seems reasonable to remove the corresponding focus areas from this wg, and let them make them evolve.


sgoggins:
Risk and value are new working groups and we agreed to merge this yesterday.

Issue #110, opened 2019-03-20
[metrics] "Reviews" instead of "Proposals"?


jgbarah:
Some people have pointed out that the name "Proposals" is difficult to map to pull requests, merge requests or code reviews, and that "proposals" is not used in this context.

Proposals came from "proposals of changes to the code", but I agree that the word "proposals", isolated, is difficult to map to a code review process. It is the name we're currently used in the code development focus area (file [code_development.md](https://github.com/chaoss/wg-gmd/blob/master/focus_areas/code_development.md)).

Thus, I propose to change "Proposals" for "Reviews", which would be much more explicit, and would still include pull requests in the case of GitHub, merge requests in the case of GitLab, and code reviews in the case of Gerrit.

The change would be in all metrics names, such as "Proposals" -> "Reviews", or "Proposals_Accepted" -> "Reviews_Accepted".

If nobody opposes, I will produce the pull request for that (or anyone who has some time could do it ,-) ).

Any feedback is welcome!!

germonprez:
This makes sense to me. I didn't even understand 'proposals' until you explained it here. Reviews is something I understand more :)

rsanchez-seerene:
Very important concept, which reminds me the document control system and RFC processes: Submit fro review, Pending review, Pending approval, release, obsoleted. However in this case the author can decline review comments/suggestions.   
![ReviewProcess](https://user-images.githubusercontent.com/19904348/54718873-d2aa6400-4b5b-11e9-9f9c-e702bfe5f404.png)

germonprez:
So it seems like the inclination is for 'review'

jgbarah:
Good. Thanks for the feedback. If there is no comments against, I will produce the pr changing this.

sgoggins:
Pull_Merge_Request? 
Forked_code_merge_request?
I think the word "review" is confusing with regards to code reviews.

jgbarah:
Thanks for the feedback, @sgoggins. My comments on it:

* Pull_Merge_Request: a bit longish, but also misses important systems, such as Gerrit, where the name has nothing to do with "pull request" or "merge request".
* Forked_code_merge_request: not sure what "forked code" means in this context...

But, from a general point of view, why "review" is confusing when talking about code *reviews*? Maybe "Code_Review" would be more clear?

jgbarah:
Any other comment about this?

sgoggins:
what about "code_integration"?

germonprez:
As related now to #119

Issue #101, opened 2019-03-06
Maintain a dictionary file of metrics and historical names


sgoggins:
For any new metric, add the name and definition to a metrics.dictionary.md file. **include any historical names that were renamed**

klumb:
I think we need to collect some basic terminology to start.

sgoggins:
Lets make it simply changes from here forward.

klumb:
add proposals and reviews

jgbarah:
I've added it as a part of #119, following the suggestion by @sgoggins.

harshalmittal4:
@jgbarah, @sgoggins, @klumb 
if there is anything else that needs to be added to name_changes.md (not sure whether to add the names of the existing metrics), let me know and I'll update it there.

jgbarah:
I think we can close this now, after merging #119 which includes a first version of that dictionary.

PR #119, opened 2019-03-27
Change "proposals" to "reviews"

jgbarah:
The name "reviews" seems much easier to understand that "proposals", so we're changing all occurrences of "proposals" to "reviews".

Closes: #110


sgoggins:
I think this is potentially confusing. See comments on the related issue.

jgbarah:
After the discussion we had during our last weekly meeting, we decided to accept the change for now. The recently uploaded new version of the patch solves conflicts, and adds a file for tracking changes in metrics names.

GeorgLink:
closes #110

GeorgLink:
closes #101

Issue #83, opened 2019-02-12
Update Readme.md headings to match standard for CHAOSS


sgoggins:
https://docs.google.com/document/d/1pVtYHNTwrGFV3UyiR2qBKV0MvipoEEI-fbwavSNui4Q/edit

Proposed Structure Revision: 

Introduction--Overview (purpose, goal, outcome, process - a sentence each with a link to the full section)
Table of Contents
Background (Problem Statement, Opportunity, Goal, Work to Date)
Key Focuses / Roadmap for 2019 [NOT IN G-M-D]
Metrics / Metrics Focus Areas
Workflow: How to use this repository (Usage) [Added from GMD ReadMe]
Implementations [Added from GMD ReadMe]
Software Work [Not included in GMD ReadMe]
Use Cases [Added from GMD ReadMe]
Related Works [Not included in GMD ReadMe]
Contributing [Moved up in sequence, as compared to GMD ReadMe]
Communications -- Meetings, Mailing List, Discussion Forum [Consolidated, as compared to GMD ReadMe]
Contributors -- Maintainers, Core Contributors
License [Added from GMD ReadMe] (related issue #149)

sgoggins:
Created this issue following discussion in weekly call.

jgbarah:
Can anyone have a look at this one, and maybe produce a pull request? In principle, it is a matter of adding the sections that could be missing, using (when necessary) the file in the D&I working group as a template. In a quick check, it is mainly the section "key focuses/roadmap" (which could just include a link to our roadmap file), and maybe "software work" (not sure about this one).

Before doing the pr, an analysis in a comment to this ticket, about the changes to do (by comparing with the D&I file, and the file linked in the description of this ticket) would be great (so that we can have a preliminary discussion, if needed).

harshalmittal4:
Hey @jgbarah, I will take it and make the PR!

GeorgLink:
D&I has not changed the headings yet. Please find new headings in the
Google doc shared on the mailing list a few weeks ago.

>

harshalmittal4:
@jgbarah, @sgoggins, @GeorgLink, 
Few questions regarding the update:
1) What could be the appropriate content for the **Background** heading for the GMD working group, or shall I just put the heading for now?
2) For the **Roadmap**, **Implementations** and **Use cases** headings, I suppose that I have to include the links to their files. Please correct me if I am wrong.
3) What would comprise the discussion forum if it is to included, in the **Communications** heading.
[...]

PR #132, opened 2019-04-04
Update README.md to match standards for CHAOSS

harshalmittal4:
Addresses #83.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 CHAOSS
+Copyright (c) 2019 CHAOSS
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019 CHAOSS
+Copyright (c) 2018 CHAOSS
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 _Growth, Maturity and Decline_ is a working group focused on the lifecycle of open source projects. We do not valence the stages, but seek to build awareness of those stages among consumers of these metrics. Decline can be ok when a project's utility is waning; or terrible if 20% of the entire internet relies on that project. Context matters, and we are not the arbiters of your context. The "valence" of "good/bad" emerges from use cases and YOUR context.
 
 ### Goal
-Currently, the goals are structured in the following focus areas : [Issue Resolution](./focus_areas/issue_resolution.md), [Code Development](./focus_areas/code_development.md) and [Community Growth](./focus_areas/community_growth.md).
+Currently, the goals are structured in the following focus areas : [Issue Resolution](focus_areas/issue_resolution.md), [Code Development](focus_areas/code_development.md) and [Community Growth](focus_areas/community_growth.md).
 
 ### Outcome
 
@@ -21,13 +21,20 @@ For each area of interest, we're following the [goal-question-metric](https://en
 ## Table of Contents
 
 - [Introduction](#introduction)
-- [Contributing](#contributing)
+<!---- [Background](#background)--->
+- [Roadmap for 2019](#roadmap-for-2019)
 - [Metrics](#metrics)
 - [Usage](#usage)
-- [Maintainers](#maintainers)
+- [Implementations](#implementations)
+<!---- [Software work](#software-work)--->
+- [Use Cases](#use-cases)
+<!---- [Related Works](#related-works)--->
+- [Contributing](#contributing)
+- [Communications](#communications)
+- [Contributors](#contributors)
 - [License](#license)
 
-## Background
+<!---## Background
 
 ### Problem Statement
 
@@ -39,20 +46,20 @@ To be updated
 
 ### Work To Date
 
-To be updated.
+To be updated. --->
 
 ## Roadmap for 2019
 
-Our key focus areas for 2019 are summarized [here](./ROADMAP.md).
+Our key focus areas for 2019 are summarized [here](ROADMAP.md).
 
 ## Metrics
 
 The metrics dealt with in this focus group are organized in focus areas:
 
 * Growth Maturity and Decline
-  * [Code Development](./focus_areas/code_development.md)
-  * [Community Growth](./focus_areas/community_growth.md)
-  * [Issue Resolution](./focus_areas/issue_resolution.md)
+  * [Code Development](focus_areas/code_development.md)
+  * [Community Growth](focus_areas/community_growth.md)
+  * [Issue Resolution](focus_areas/issue_resolution.md)
 
 ## Usage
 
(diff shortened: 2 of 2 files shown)


harshalmittal4:
@jgbarah  @sgoggins @GeorgLink,
Please review for the required changes if any.

harshalmittal4:
@jgbarah, @valeriocos ping!

README.md
```diff
@@ -60,26 +120,8 @@ so that anyone had the opportunity of commenting on them, and make up their mind
...
-This is a repository for the GMD working group.
-When this working group decides to "release" a version of its metrics,
-the relevant files will be dumped to the
-[chaoss/metrics](https://github.com/chaoss/metrics) repository,
-as a pull request for updating the CHAOSS metrics with
-"GMD metrics release x.y".
-
 
-## Maintainers
+## Contributors
```
valeriocos: Not sure if `Maintainers` should be replaced. What do you think @jgbarah ?

README.md
```diff
@@ -4,44 +4,104 @@
...
+## Roadmap for 2019
+
+Our key focus areas for 2019 are summarized [here](./ROADMAP.md).
+
+## Metrics
+
+The metrics dealt with in this focus group are organized in focus areas:
+
+* Growth Maturity and Decline
+  * [Code Development](./focus_areas/code_development.md)
```
valeriocos: I'm not sure if it's better to put the full link instead of the dot (https://github.com/chaoss/wg-gmd/blob/master/focus_areas/code_development.md instead of ./focus_areas/code_development.md)

valeriocos:
@jgbarah I had a look at the PR and LGTM, however your feedback is needed to be sure that everything is OK.

README.md
harshalmittal4: Actually `Contributors` was mentioned first in the strucure proposed, so did accordingly. But  `Maintainers` is also mentioned there. Sorry :sweat_smile: I also feel that `Maintainers` would be more appropriate. 
Thanks! Let me know @valeriocos.
@jgbarah ^ Please see!

README.md
harshalmittal4: I would rectify the PR to include the complete links!

README.md
jgbarah: We have both: maintainers are those with write permission in the repo, contributors are (if I'm not wrong) any other contributors.

README.md
jgbarah: > I'm not sure if it's better to put the full link instead of the dot (https://github.com/chaoss/wg-gmd/blob/master/focus_areas/code_development.md instead of ./focus_areas/code_development.md)
> 
> This comment applies to most links in the PR

I think it is better to have the relative link, since that way it will always point to the right place, even if the repo changes in name (which is something that could happen during the next few days). Additionally, I would skip `./`, since the relative link would work also without it.

harshalmittal4: Ok. @jgbarah, will do the needful!

README.md
harshalmittal4: So what heading should appear in the README, can it both, something like:
 `Contributors / Maintainers`

harshalmittal4: @valeriocos, @jgbarah ping!

harshalmittal4: @jgbarah, please have a look at this. What sould come then 

> We have both: maintainers are those with write permission in the repo, contributors are (if I'm not wrong) any other contributors.
@jgbarah, just a gentle reminder, please have a look as you get time. Let me know what would be appropriate in the README.md then. Thanks!

GeorgLink:
fixes #83

LICENSE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2018 CHAOSS
+Copyright (c) 2019 CHAOSS
```
GeorgLink: Please make pull requests for one thing at a time. 

For example, do not change the license file, when the pull request is about updating the README.

README.md
GeorgLink: @harshalmittal4 you can take a look at the D&I WG. The Contributors section has a Maintainers subsection, a Core Contributor subsection and includes a list of all contributors. Maybe that is a good solution? 

https://github.com/chaoss/wg-diversity-inclusion

LICENSE
harshalmittal4: Sure @GeorgLink, I would make a new PR for the license file including the previous year!

README.md
harshalmittal4: Yes this seems to be a good solution, just one thing, can you please tell what all names should I include in the `Core Contributors` section. Or as of now shall I leave the `Core Contributors` header vacant?

LICENSE
harshalmittal4: Sorry about this @GeorgLink, will make a seperate PR for this change.

GeorgLink: Thanks @harshalmittal4 
With pull request #135, the LICENSE file change can be removed form pull request #132 (this pr)

README.md
GeorgLink: I don't know for GMD who the core contributors are. Maybe @sgoggins or @jgbarah know. If you don't get names, remove the section for now.

LICENSE
harshalmittal4: Yes, @GeorgLink, but I first need to resolve the conversation corresponding to the Maintainers/Contributors to update this PR. Please update me on it :)

harshalmittal4:
Hey @GeorgLink, please review, does this look good now.

GeorgLink:
Thanks for fixing the LICENSE file change.

In the README, how about filling in or removing empty sections? The information `To be updated.` is not valuable by itself. A better way might be to create an issue for each of these sections and add them to the readme only after content was developed.

harshalmittal4:
Hey @GeorgLink, I commented out the empty sections instead of removing, to keep track of adding them in the same order once the content is developed. Is this fine.

GeorgLink:
> Hey @GeorgLink, I think it is better to comment out the empty sections instead of removing them, to keep track of adding them in the same order once the content is developed. Is this fine or shall I modify the PR to remove the comments.

I'd be okay with it. I defer this decision to the maintainers of the repository. 

Thank you for continuing to work on this.

harshalmittal4:
@valeriocos, @jgbarah, review required as you get time :smile:

valeriocos:
lgtm! thank you @harshalmittal4

README.md
jgbarah: > can you please tell what all names should I include in the `Core Contributors` section. Or as of now shall I leave the `Core Contributors` header vacant?

AFAIK, we don't have a difference except for maintainers and contributors. If the name of maintainers and/or contributors get high enough, we could discuss who is a core contributor (and what does it mean), but for now, I would remove that one.

jgbarah:
To be honest, I'm no longer sure what we're trying with this pull request. Comparing with the D&I README, I see sections that are the same, and sections that are in one of them, but not in the other. In the end, maybe there is a tension between the [Standard Readme](https://github.com/RichardLitt/standard-readme) and the one in D&I, which only partially resembles it.

To avoid further discussion, and have a useful README, i suggest accepting this pr (because I think we're starting to get lost in the details), and then, in a separate pr, make it closer to the D&I README, to have some more uniform READMEs in CHAOSS.

Please, comment on this, or just merge this pr and we start working in the new one.

harshalmittal4:
Hey @jgbarah, this PR tries to follow the structure proposed in #83, omitting the sections for which content is not present for now (I have commented out those headings in Readme's markdown). I checked the D&I Working Group's Readme, it also tries to the adhere to the same structure except the fact that some sections are missing there too. Maybe's thats why some non-uniformity.

jgbarah:
Thanks for the clarification, @harshalmittal4. Yes, I know about the proposal in #83, and I'm happy with it. But I have the feeling that maybe we're now in the middle between that proposal, the D&I Readme, and the original proposal on following some standard for Readme files...

But in any case, as I commented, I think we can just accept this pr, and follow on from here...

Thanks a lot for moving this forward, @harshalmittal4

harshalmittal4:
Yes sure @jgbarah whatever you feel good :)

PR #143, opened 2019-04-24
Display Focus Area overview as a table

GeorgLink:
The idea of this pull request is to display focus areas as a table with a description of the focus areas. This is supposed to help someone coming to the repository to know what each focus area is about.

Further, this pull request replicates the table in a README for the /focus-area dirctory, so that it is more user friendly when someone looks through the structure of the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,13 @@ Our key focus areas for 2019 are summarized [here](ROADMAP.md).
 
 ## Metrics
 
-The metrics dealt with in this focus group are organized in focus areas:
+The evolution metrics dealt with in this focus group are organized in focus areas:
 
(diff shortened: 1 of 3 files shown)


GeorgLink:
> Concerning proposals, is there a time frame to consider? For example, we mention ... during a certain time period, which is?

Sorry @foundjem, I don't understand your question. Are you thinking a focus area needs to be described differently? Maybe a different issue or pull request would be good for that.

This pull request is not about framing focus areas but about how they are presented in the README.

GeorgLink:
Fixed merge conflict.

Odd. I don't mean to edit the code_development.md file, only the README.md.

PR #158, opened 2019-06-04
Include reference to CoC instead of CoC itself

jgbarah:
Following the discussion during the monthly CHAOSS meeting today,
I'm substituring the whole CoC by a reference to the "canonical" CHAOSS CoC.


foundjem:
Sounds good to me.

PR #162, opened 2019-06-10
[GSoC] Add reference implementation and python script for CodeChanges metric

### LINKED ISSUE #153: [GSoC] Re-implementing the Code_changes-git metric
As discussed with @jgbarah and @aswanipranjal in the IRC meeting on the 24th of May (log can be found [here](https://github.com/Polaris000/GSoC_19_Perceval_Implementations/blob/master/IRC_chats/may_24_comm_bonding.txt), and blog post [here](https://polaris000.github.io/blog/com_bonding_2)), a new structure for reference implementations is being tried out:  

Create a class for the analysis of reference implementations with several methods, the most important being a method which would calculate a single value for the metric. Some other
```python
  class MetricCompute(Period, Source_Code_criteria):

      # generate a value for the metric
      method1: compute(self, data) 

      # generate a value for each period: like one per week/month
      method2: compute_time_series(self, data, period)

      # generate charts / graphs regarding the metric
      optional: generate_graphs(self, data, ..)
```
methods can be added depending on the metric to create charts or graphs. This notebook will be used to generate a script, which would be tested against some sample Perceval JSON data. Then a user can import this as a module to run it on their own data. 

Since this method is still under evaluation, it was decided that I would implement this metric using the new method in my [project-tracker/experiment repo](https://github.com/Polaris000/GSoC_19_Perceval_Implementations). If this proves to be effective (and I am sure it will :) ), other metrics will  be implemented this way. 

Please drop your ideas and suggestions below!

### LINKED ISSUE #154: [GSoC] A root class from which all reference implementations inherit functionality
My mentors (@jgbarah @valeriocos and @aswanipranjal) and I felt that creating a parent class for the entire module of python scripts generated from the reference implementations would be a great idea. This was discussed in the latest weekly IRC meeting. This class will have all of the existing reusable functionality, like reading data from Perceval and converting it to a Pandas DataFrame. This is part of my tasks for the coming week.

All are welcome to share their views on this idea!

### LINKED ISSUE #155: [GSoC] Update the Code_changes-git implementation using a root class
Following #154 and #153, one of my tasks for this week is to create a reference implementation for Code_changes-git, keeping both the ideas (discussed in the issues) in mind. 
As of now, this will be present in my [project tracker repository](https://github.com/Polaris000/GSoC_19_Perceval_Implementations). Once, it is approved, it will be added to the Evolution working group repository.

Polaris000:
- Closes #153 
- Closes #155.
- Closes #154


Polaris000:
Please have a look at @jgbarah!

Polaris000:
After following the last few comments in #160, please have a look at this PR. Thanks!

implementations/scripts/Metric.py
```diff
@@ -0,0 +1,41 @@
+import pandas as pd
+
+
+class Metric():
+    """
+    Create a dataframe from data-list, which is the data
+    fetched by Perceval
+
+    :param data_list: A list of dictionaries.
```
jgbarah: Please, use "items" instead of "data_list". See rationale in other comment in this review.

implementations/scripts/Metric.py
```diff
@@ -0,0 +1,41 @@
...
+        Each element is a Perceval dictionary, obtained from a JSON
+        file or from Perceval directly.
+    """
+
+    def __init__(self, data_list):
+        data_list_flattened = Metric._flatten_data(data_list)
+        self.raw_df = pd.DataFrame(data_list_flattened)
```
jgbarah: `data_list_flattened` -> `flat_items`, which I think is easier to understand.

implementations/scripts/Metric.py
```diff
@@ -0,0 +1,41 @@
...
+    Create a dataframe from data-list, which is the data
+    fetched by Perceval
+
+    :param data_list: A list of dictionaries.
+        Each element is a Perceval dictionary, obtained from a JSON
+        file or from Perceval directly.
+    """
+
+    def __init__(self, data_list):
+        data_list_flattened = Metric._flatten_data(data_list)
```
jgbarah: This should be extensible by child classes, and therefore, `self._flatten_data()` (or maybe `class._flatten_data()`)

implementations/scripts/Metric.py
```diff
@@ -0,0 +1,41 @@
...
+        self.raw_df = pd.DataFrame(data_list_flattened)
+
+    @staticmethod
+    def _flatten_data(data_list):
```
jgbarah: I'm not sure we're on the same page with respect to this method. As i commented above, I think this should be a function provided by children classes. For example, `Commit` (or maybe `Code_Changes`) will define it to get the fields from the Perceval dictionary that you really need to produce the flat directory that you need to produce a row in the dataframe. You don't really need all fields in the original Perceval directory, only some of them. `Commit` or maybe `Code_Changes` would know which fields are actually needed for implementing the metric and its filters.

implementations/scripts/Commit.py
```diff
@@ -0,0 +1,90 @@
...
+        It is of the form (since, until), where since and until are strings.
+        Either, or both can be None. If, for example, since is None, that 
+        would mean that all commits from the first commit to the commit who last
+        falls inside the until range will be included.
+
+    :param is_src_code_obj: An object of IsSourceCode.
+        It is used to determine what comprises source code.
+        """
+
+    def __init__(self, data_list, date_range=(None, None), is_src_code_obj=None):
```
jgbarah: Do you think you really need to redefine this method for all children? I think the functionality is pretty much the same for all metrics, so I would include most of what you do here in the 'Metric' class. I'm only not sure about `date_range`, but I think all metrics we're defining work only for a date range, so it could go to Metric.

implementations/scripts/Commit.py
```diff
@@ -0,0 +1,90 @@
...
+        else:
+            self.since = utils.get_date(self.df, "since")
+
+        if self.until:
+            self.df = self.df[self.df['created_date']
+                              < utils.str_to_dt_other(self.until)]
+        else:
+            self.until = utils.get_date(self.df, "until")
+
+    def _clean_commit(self, line):
```
jgbarah: I think this is mainly the `_flatten_data` that I comment elsewhere that should be defined in the child classes (such as this one)

implementations/scripts/Commit.py
```diff
@@ -0,0 +1,90 @@
...
+
+        super().__init__(data_list)
+
+        clean_data_list = list()
+        (self.since, self.until) = date_range
+
+        for line in self.raw_df.iterrows():
+            commit = line[1].to_dict()
+            commit = self._clean_commit(commit)
+            if is_src_code_obj is None  \
```
jgbarah: We need to find a way of doing this in the Metric class, so that we don't need to call `__init__` for the parent class (except that it is too cumbersome). We could have it as an argument when instantiating the class, and just storing it eg self._is_source, so that we can call it when needed.

implementations/scripts/Commit.py
```diff
@@ -0,0 +1,90 @@
...
+                    or is_src_code_obj.is_source_code(commit):
+                clean_data_list.append(commit)
+
+        self.df = pd.DataFrame(clean_data_list)
+        if self.since:
+            self.df = self.df[self.df['created_date']
+                              >= utils.str_to_dt_other(self.since)]
```
jgbarah: Since we have datetime in Python for dates, we should just use it for since / until, and for dates coming from Perceval.

In other words, let's make all our classes work with datetime, so that people can use the usual Python conversions (if needed) when using them. For that, the functioning flattening data from Perceval would be in charge of converting dates, as produced by Perceval, to datetime.

implementations/scripts/SourceCode.py
```diff
@@ -0,0 +1,101 @@
...
+        :param method: a string which can be "naive",
+            "folder_exclude" or "extension_exclude"
+        For example,
+        source_code_exclude_list = ["py", "other", "gitignore"]
+        or source_code_exclude_list = ["tests/", "bin/", "drivers/base/"]
+        """
+        self.source_code_exclude_list = source_code_exclude_list
+        self.method = method
+
+    def is_source_code(self, commit):
```
jgbarah: Since the class is already `IsSourceCode`, this can be, for example, `check()`.

implementations/scripts/SourceCode.py
```diff
@@ -0,0 +1,101 @@
+class IsSourceCode:
```
jgbarah: If you want to implement this as a class, that's fine with me. But it should be a hierarchy of classes. You can start for example with a naive class (that let's anything be source code), but then you can extend from it (or just provide another class with the same `check()` function that implements for example the heuristics you have here, based on that list of excluded paths. Then, when instantiating the Metric, you would provide as argument one of these classes (or objects made from it).

implementations/scripts/SourceCode.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        Given a commit structure, which is a dictionary returned
+        by the _summary function, and given a list of files to exclude
+        using source_code_exclude_list while instantiating an object,
+        decide whether all the files in a commit are to be excluded or not.
+
+        :param commit: a commit structure, returned by the _summary method.
+        """
+        options_dict = {
+            "naive": self._naive_implementation,
```
jgbarah: I think each of these options would be the semantics of one of those classes that I mention above.

implementations/scripts/utils.py
```diff
@@ -0,0 +1,63 @@
+import datetime
+import json
+
+
+def str_to_dt_other(date):
```
jgbarah: I think this is not needed, see above where you use this function, and the discussion on `datetime`.

implementations/scripts/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    except ValueError as ve:
+        datetimestr = datetime.datetime.strptime(
+                date, "%Y-%m-%dT%H:%M:%SZ").strftime("%Y-%m-%d")
+
+    finally:
+        datetimeobj = datetime.datetime.strptime(datetimestr, "%Y-%m-%d")
+        return datetimeobj
+
+
+def get_date(df, option="since"):
```
jgbarah: I'm not sure this is really needed, but we can discuss later, during the next iteration.

Polaris000: The reason I added this was for the `compute_timeseries` method: I needed a way to make sure that the dataframe returned was between a definite range. If no `since` or `until` are provided, I made it so that the earliest event in the JSON data would become `since` and the latest would be `until`.

jgbarah: Yes, I understood that, but maybe the timeseries can be constructed without this. However, leave it this way, if you want. We can enter in details during the next iteration.

implementations/scripts/SourceCode.py
Polaris000: Got it! will do.

implementations/scripts/SourceCode.py
Polaris000: Oops! will change that.

implementations/scripts/Commit.py
Polaris000: Got it!

implementations/scripts/Metric.py
Polaris000: OK

implementations/README.md
```diff
@@ -1,9 +1,30 @@
 # Examples of implementations for metrics
 
 Notebooks, built on top of data collected by Perceval,
-for illustrating how to compute metrics:
+for illustrating how to compute different metrics of the Evolution Working Group. The notebooks will be explanatory in regard to the analysis performed. They should help the user understand how to calculate these metrics, as well as to understand how the metric is defined. 
+  
+The contents of this directory are the notebooks, as well as some scripts in the scripts directory. The scripts directory is a package containing the following modules:  
 
-* [Code Commits](Code_Commits.ipynb),
+- **Metric.py**:
```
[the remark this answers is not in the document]
Polaris000: Done!

implementations/README.md
```diff
@@ -1,9 +1,30 @@
...
+- **Individual Metric classes**:  
+These are python scripts converted from the reference implementation notebooks. They inherit from the Category modules, but have the most important function of all --- to calculate values for the metrics they represent.   
+
+To summarize, the structure of the implementations is like so:
+```
+Root class (Metric.py) <- Category classes (Commit.py, for example) <- Metric classes
+```
+
+### Note
+- To make it easier to understand these implementations, one basic assumption is that each metric is computed for all data in the file, and currently, not for each repository. Remember that the JSON file analyzed can have data from several different sources appended to it.
```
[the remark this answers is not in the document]
Polaris000: That is what I had tried to explain. Obviously, the point was a bit vague. Have updated it. Thanks!

implementations/README.md
```diff
@@ -1,9 +1,30 @@
...
+These are python scripts converted from the reference implementation notebooks. They inherit from the Category modules, but have the most important function of all --- to calculate values for the metrics they represent.   
+
+To summarize, the structure of the implementations is like so:
+```
+Root class (Metric.py) <- Category classes (Commit.py, for example) <- Metric classes
+```
+
+### Note
+- To make it easier to understand these implementations, one basic assumption is that each metric is computed for all data in the file, and currently, not for each repository. Remember that the JSON file analyzed can have data from several different sources appended to it. 
+- Another assumption is that the JSON data file has data points of the same category, that is, all of them must belong to exactly one of the commit, issue or pull_request categories.
```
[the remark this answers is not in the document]
Polaris000: Done!

implementations/README.md
```diff
@@ -1,9 +1,30 @@
...
+
+## A list of implementations
+* [Code Changes](Code_Changes_Git.ipynb),
```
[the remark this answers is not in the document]
Polaris000: Done!

implementations/README.md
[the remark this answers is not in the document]
Polaris000: Screenshot from 2019-06-13 12-17-18

Do the links above look OK , or do you suggest doing: 
```<module_name> (<module_link>, <notebook_link>)```

File: implementations/scripts/utils.py [hunk 18]
```diff
@@ -0,0 +1,63 @@
...
+
+
+def read_JSON_file(path):
+    """
+    Given a line-by-line JSON file, this function converts it to
+    a Python dict and returns all such lines as a list.
+
+    :param path: the path to the JSON file
+    """
+    data_list = list()
```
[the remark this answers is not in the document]
Polaris000: Done

implementations/scripts/Metric.py
Polaris000: Done!

implementations/scripts/SourceCode.py
Polaris000: I could try implementing a simple factory pattern? 
Something like this (names are only for explanatory purposes):
```
class Option():
    check()

class Naive(Option):
    check()

class FileExclude(Option):
    check()
...
..

class ChooseOption():
    # choose right option
    option.check()
```

implementations/scripts/Metric.py
Polaris000: I agree with what you say. But if `_flatten_data` is defined in `Commit`, where it will directly act on Perceval data and flatten it while keeping only those fields which are required, what is the need for the root Metric class? This is where I am stuck.
Do you want it so that the Metric(root) class is never explicitly called and just acts like a template of each of its children (the commit, issue and pull request classes)?
The way I am using these classes as of now is to create a object for every individual metric class (like CodeChanges, for example), and each class calls its parent's `__init__`

Please let me know. 
As of now, we have three main "checks":
- date range  
- source code  
- keeping only required fields.

**date range** is common to each and every metric we implement, irrespective of category (commit, issue, pull request).
**source code**: as of now, I can't think of a way to check for issues (but we will see that later)
**keeping only required fields**: This is specific to each Category.

implementations/scripts/Commit.py
Polaris000: Please see my comment above

implementations/README.md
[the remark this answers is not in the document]
Polaris000: Also, do you want to have notebooks for all non_metric(Metric(root), Commit,...) modules too, and have them added to `notebooks_df` (for pandas implementations)?

implementations/scripts/Commit.py
Polaris000: Please have a look at my comments above. I have a few queries regarding this part. Thanks!

implementations/scripts/Commit.py
Polaris000: Please have a look at my comments above.

Polaris000:
@jgbarah, please have a look at the new commits as well my comments in the review you performed yesterday. Once, this is resolved, I'll squash the commits / add a proper discription. 
Thanks!

implementations/README.md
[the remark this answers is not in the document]
jgbarah: > do you want to have notebooks for all non_metric(Metric(root), Commit,...) modules too

I think that's a good idea, to have stuff more consistent (and easy to follow).

implementations/scripts/SourceCode.py
jgbarah: Yes, something like that, except for the `ChooseOption` class, which I don't understand. Usually, you would just pass the intended class as a parameter when instantiating the corresponding object in the Metrics hierarchy, and that object will just call check in that class for each commit...

Polaris000: If you look at my latest patch:
```python
class IsSourceCode:

     def __init__(self, source_code_exclude_list=None, algorithm_class=None):
        self.algorithm_obj = globals()[algorithm_class](source_code_exclude_list)

     def check(self, commit):
        """
        Returns the check method of the algorithm to be used. 
        :param commit: a commit structure, returned by the _flatten_data method.
        """
        return self.algorithm_obj.check(commit)
```
it allows the user to enter a string and not have to create an object and pass it to the function. The algorithm classes are hidden from the user, so to speak.

Polaris000: Also, please have a look at my comment [here](https://github.com/chaoss/wg-evolution/pull/162#discussion_r293298796): I have a doubt regarding the `flatten` method.

implementations/README.md
```diff
@@ -1,11 +1,36 @@
...
+- **Individual Metric modules**:  
+These are python scripts converted from the reference implementation notebooks. They inherit from the Category modules, but have the most important function of all --- to calculate values for the metrics they represent.   
+
+To summarize, the structure of the implementations is like so:
+```
+Root class (metric.py) <- Category classes (commit.py, for example) <- Metric classes
+```
+
+### Note
+- Every metric is calculated for the JSON file as a whole. If a JSON file has the data of only one repository, only that will considered for calculating the metric. Likewise, using a JSON file with the data of an entire project,
```
jgbarah: But metrics are not actually computed on JSON files, but on data structures with items... We read the JSON file when showing how to use the metric...

Polaris000: OK, will make it clearer. Regarding the html file, I forgot to remove it. Will do so!

implementations/code_df/code_changes_git.py
```diff
@@ -0,0 +1,154 @@
...
+        interest
+    :param is_src_code_obj: An object of IsSourceCode, to be used to determine
+        what comprises source code.
+    """
+
+    def __init__(self, items, date_range=(None, None), is_src_code_obj=None):
+
+        super().__init__(items, date_range, is_src_code_obj)
+
+    def total_count(self):
```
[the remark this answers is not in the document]
Polaris000: oh.. its there in the existing implementation. I 'll remove it if not required.

implementations/code_df/code_changes_git.py
```diff
@@ -0,0 +1,154 @@
...
+    def _count_master_only(self, incl_empty=True):
+        """
+        Counts commits present only on the master branch.
+
+        :param incl_empty: exclude empty commits on the master branch
+        """
+
+        df = self.df
+        todo = set()
+        for _, commit in df.iterrows():
```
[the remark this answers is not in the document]
Polaris000: Will have a look.

implementations/code_df/code_changes_git.py
```diff
@@ -0,0 +1,154 @@
...
+        """Count number of commits of different types, like including empty commits
+        or counting only those commits made on the master branch.
+
+        :param incl_empty: Include empty commits
+        :param incl_merge: Include merge commits
+        :param master_only: Include only commits made on master branch
+        """
+        df = self.df
+
+        if master_only:
```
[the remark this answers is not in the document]
Polaris000: :) of course!

implementations/code_df/code_changes_git.py
```diff
@@ -0,0 +1,154 @@
...
+            all_periods = pd.DataFrame(
+                [all_periods['Dates'].dt.year.rename("year"),
+                 all_periods['Dates'].dt.week.rename("week")]).T
+
+        timeseries_df = pd.DataFrame(timeseries_series)
+        timeseries_df.reset_index(inplace=True)
+        timeseries_df.columns = ["year", period, "count"]
+        merged_df = all_periods.merge(timeseries_df, how='outer').fillna(0)
+
+        if plot_chart:
```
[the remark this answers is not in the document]
Polaris000: I am removing the plotting part for now. Over the weekend, I'll try and come up with some simple and modular design with classes.

implementations/code_df/code_changes_git.py
```diff
@@ -0,0 +1,154 @@
...
+        JSON file
+    :param date_range: A tuple which represents the start and end date of
+        interest
+    :param is_src_code_obj: An object of IsSourceCode, to be used to determine
+        what comprises source code.
+    """
+
+    def __init__(self, items, date_range=(None, None), is_src_code_obj=None):
+
+        super().__init__(items, date_range, is_src_code_obj)
```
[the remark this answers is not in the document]
Polaris000: Could you please elaborate? `items` is present as a parameter to both __init__ methods. It  has also been mentioned in the docstring...

implementations/code_df/code_changes_git.py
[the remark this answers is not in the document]
Polaris000: Let me know if that is OK

Polaris000:
@jgbarah, unfortunately I am running into some bugs after the recent changes. As it is getting late here (IST), I'll only be able to send the next patch tomorrow at the earliest. I am really sorry. 

I had a few queries (after latest meet). Though they are minor, I'll ask them here so we won't have to review them later in the next iteration:
1. IsSourceCode:
So my current structure (the last changes I pushed to this branch and you reviewed) involves all algorithm classes inheriting from a `Algorithm` class. Since we are planning on only using classes (and not having to create objects), the `__init__` method is not of any use. 
This is purely in regard to better readability but should I just not define __init__ in any class?
Or I could do something like this: 
```python
class Naive(Algorithm):
    """
    Instatiates an object of Naive. It is one of several
    "algorithms" used to define source code. 
    """
    def __init__(self):
        super().__init__()
   
     @staticmethod
     def check().....

```
where Algorithm is:
```python
class Algorithm:
    """
    Parent class to all Algorithms.
    """
    def __init__(self):
        pass
    
    @staticmethod
    def check ... 
```
Which one do you prefer?

jgbarah:
> @jgbarah, unfortunately I am running into some bugs after the recent changes. As it is getting late here (IST), I'll only be able to send the next patch tomorrow at the earliest. I am really sorry.

Don't worry. Do it at your earliest convenience.

> I had a few queries (after latest meet). Though they are minor, I'll ask them here so we won't have to review them later in the next iteration:
> 
>     1. IsSourceCode:
>        So my current structure (the last changes I pushed to this branch and you reviewed) involves all algorithm classes inheriting from a `Algorithm` class. Since we are planning on only using classes (and not having to create objects), the `__init__` method is not of any use.
>        This is purely in regard to better readability but should I just not define **init** in any class?
>        Or I could do something like this:

I'm still not sure we really don't need objects, but for now, we can work with classes and class methods. In this case, you can define __init__, but make it raise an exception.

My main concern is that if we don't need to instantiate the class, having just functions (without being embedded in classes) could be enough, and is simpler.

But for now, let's go with classes. Just a minor comment: don't name the root class `Algorithm`, which is too generic. Something like `IsSourceCode` or `CodeCriteria` would be clearer.
   
> 
> ```python
> class Naive(Algorithm):
>     """
>     Instatiates an object of Naive. It is one of several
>     "algorithms" used to define source code. 
>     """
>     def __init__(self):
>         super().__init__()

In any case, you don't need to redefine '__init__', if it does nothing new. It will just inherit the `__init__` in the parent.
    
>      @staticmethod
>      def check().....
> ```
> 
> where Algorithm is:
> 
> ```python
> class Algorithm:
>     """
>     Parent class to all Algorithms.
>     """
>     def __init__(self):
>         pass
>     
>     @staticmethod
>     def check ... 
> ```
> 
> Which one do you prefer?

Is it more clear now?

Polaris000:
Thanks @jgbarah. I had left a few comments on your last review. Please have a look whenever convenient. Regarding your response for my query above:
- `Algorithm` is different from `IsSourceCode`. `IsSourceCode` is like a factory class, which calls the `check()` method for the class passed to it while instantiating its object
`Algorithm` is the base class for all algorithms (`Naive`, `FolderExclude`, ...)..
I'll send the patch and you can have a look at the naming.

implementations/code_df/is_source_code.py
```diff
@@ -0,0 +1,163 @@
...
+    def check(source_code_exclude_list, file):
+        raise NotImplementedError
+
+
+class Naive(Algorithm):
+    """
+    Instantiates an object of Naive. It is one of several
+    "algorithms" used to define source code.
+    """
+    def __init__(self):
```
Polaris000: @jgbarah, you mentioned raising a NotImplementedError here. Is that recommended? Would it be better to just not add __init__?

implementations/code_df/is_source_code.py
```diff
@@ -0,0 +1,163 @@
...
+        super().__init__()
```
Polaris000: @jgbarah, you mentioned raising a NotImplementedError here. Is that recommended? Would it be better to just not add __init__?

implementations/code_df/commit.py
```diff
@@ -0,0 +1,89 @@
...
+            commit = self._flatten_data(line)
+            if commit is not None:
+                clean_items.append(commit)
+
+        self.df = pd.DataFrame(clean_items)
+
+        if self.since is None:
+            self.since = utils.get_date(self.df, 'since')
+        if self.until is None:
+            self.until = utils.get_date(self.df, 'until')
```
Polaris000: @jgbarah, though I moved everything to `_flatten_data`, I had to move the `since, until` population in case they are `None` back to `__init__`, since they are to be populated with the minimum value of all commits. Later, when we modify `compute_timeseries` (to avoid this problem), I'll remove this part. 
`_flatten_data` works for each commit, and hence it does not seem proper to keep a track of min/max date among all commits and at the last iteration, assign that to `since` or `until`.

Polaris000:
@jgbarah, please have look at this patch whenever convenient. I have mentioned a few points in **my** review above for this patch.

implementations/code_df/is_source_code.py
jgbarah: I think this is a corner case, since it is very rare to have a class which is not intended to be instantiated (see my other comments about this). In any case, following the "explicit better than implicit", if the class is not intended to be instantiated, I think it is better that `__init__` raises an exception. See [documentation of NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError): although it is meant for abstract classes, it is the closer case I can think of.

Polaris000: OK, will do!

Issue #159, opened 2019-06-07
[sigils] Document how GrimoireLab dashboards can display released metrics


GeorgLink:
This issue proposes to include in the known implementation sections of metrics a description of how GrimoireLab provides these metrics. 

This requires updates to GrimoreLab Sigils dashboard description to align them with the CHAOSS metric definition.

jgbarah:
Good idea, @GeorgLink !!

Some pointers, just in case:

* [Example of a dashboard description in Sigils for a CHAOSS metric](https://chaoss.github.io/grimoirelab-sigils/chaoss-gmd-cde/lines_of_code_changed/). Maybe it would be good enough to write these kind of descriptions in Sigils, and link from the metrics description, or you think it is better to include a part of this content in the metrics description itself?

* [Live dashboard including the above metric, in the CHAOSS dashboard](https://chaoss.biterg.io/app/kibana#/dashboard/f13af0e0-18e5-11e9-ba47-d5cbef43f8d3)

* [Description page in Sigils](https://github.com/chaoss/grimoirelab-sigils/blob/master/docs/_chaoss-gmd-cde/lines_of_code_changed.md) for the same metric. The description in the dashboard, linked above, comes from this file.

GeorgLink:
Thanks @jgbarah. I don't know what the best format is. 
@alpgarcia and I will explore this together on Tuesday. 
We want to have a first draft (pull request) ready to discuss at the next Evolution WG meeting.

jgbarah:
I think we're done with this. Thanks!

PR #163, opened 2019-06-11
add code_changes_lines grimoirelab example

GeorgLink:
This pull request implements the idea from #159 and adds to the metric "Code Changes Lines" details about how to get the metric in GrimoireLab.

This pull request is an idea for how we can make the metric definitions more actionable for users who are interested in using the metric.

Please provide comments and feedback.

This pr was created with the help of @alpgarcia, @mghfdez, and @dicortazar.

/cc @sduenas @valeriocos @sgoggins @jgbarah @germonprez


GeorgLink:
Thanks @jgbarah: I fixed the DCO sign-off.

What do other WG members think about this approach of more closely linking metrics with software.

jgbarah:
> What do other WG members think about this approach of more closely linking metrics with software.

I'm requesting review by some other people, and would merge this if no negative feedback is received.

germonprez:
This came up in the June 11 CHAOSS community call. I hope that all metrics can have something like this. So I approve.

jgbarah:
Great! Merging then. Thanks to everyone!

PR #164, opened 2019-06-11
Add Reviews_Duration as a detailed metric

jgbarah:
We already accepted Reviews_Duration in a past pull request.
This patch adds it to the corresponding focus area file.


jgbarah:
Any comment on this one?

PR #171, opened 2019-06-17
Add GrimoireLab implementation of issues_active

alpgarcia:
This pull request implements the idea from #159 and adds to the metric "Issues Active" details about how to get the metric in GrimoireLab.

This pull request follows the same idea as in the previous pull request #163 and makes metric definitions more actionable for users who are interested in using the metric.

Please provide comments and feedback.

/cc @GeorgLink 

Disclaimer: I borrowed this description from previous @GeorgLink similar PRs :)


GeorgLink:
@alpgarcia I need to ask clarification on how the data works here.

Assume the following scenario.
Issue A was updated in week 1 and week 2.
My understanding of the graph is that issue A will only be counted in week 2 because that is the most recent update. Therefore, week 1 will show not # of active issues during that week because it is missing Issue A.

If my understanding is correct, then we can resolve the misunderstanding in one of two ways:
1. Don't show over time, but only show a single number.
2. Don't call it `# Issues Active` but `# Issues Last Active` to indicate that the graph only shows when issues were last active, not all weeks they were active in.

alpgarcia:
@GeorgLink you are correct, that's a limitation of GrimoireLab data. Thanks a lot for the clear explanation!

Probably 1 is the best option. If we try to find an intersection between a general definition of activity like: `having any update within a given time period` and the data we have, i.e.: `last update`, we could end with something like: `Issues last updated from a given date to now`. Adding `now` is key because that makes last update valid as an indicator of activity. Using any previous end date will end with the problem you pointed out.

Does that make sense to you?

GeorgLink:
@jgbarah I agree that GrimoireLab does not fully support this metric yet. However, in the dashboard, you can specify a time frame and have a single number returned with issues that were last active during that timeframe. That is a very closely related metric, assuming that the timeframe ends on the current date.

I support @alpgarcia's suggestion to acknowledge the limitation but still including a Single Metric visualization for "issues active from x to now"!

foundjem:
LGTM

> > I agree that GrimoireLab does not fully support this metric yet. However, in the dashboard, you can specify a time frame and have a single number returned with issues that were last active during that timeframe. That is a very closely related metric, assuming that the timeframe ends on the current date.
> 
> OK, i understand. in that case, please the text that explain that the metric is close, but not exactly the one defined, so that there is no confusion. I agree that the visualization is still useful.
> 
> So, from my side, once that clarification is included, I'm good with the patch.

LGTM, however, instead of only acknowledging GrimoireLab’s limitation, we can add the missing features since it has already been tested/used in the field.

jgbarah:
> instead of only acknowledging GrimoireLab’s limitation, we can add the missing features since it has already been tested/used in the field.

What do you mean? Adding the missing features to GrimoireLab?

foundjem:
> > instead of only acknowledging GrimoireLab’s limitation, we can add the missing features since it has already been tested/used in the field.
> 
> What do you mean? Adding the missing features to GrimoireLab?

Yes, if that is possible it will be better. Since GrimoireLab has wide usage thus far.

PR #172, opened 2019-06-17
[GSoC] Add reference implementation for new_contributors_of_commits

### LINKED ISSUE #157: [GSoC] Create implementations for a few more metrics
After finishing the Code_changes-git metric reference implementation, my goal is to finish a few more this week. The metrics to be worked on will most likely be all other commit related metrics.

Polaris000:
@jgbarah:
 Please have a look whenever convenient. :)

- Closes #157


Polaris000:
Closes #157

jgbarah:
Please, check [my comment](https://github.com/chaoss/wg-evolution/issues/157#issuecomment-503335939) about the new structure in #177.

Polaris000:
Please have a look @jgbarah.

implementations/code_df/new_contributors_of_commits.py
```diff
@@ -0,0 +1,147 @@
...
+
+class NewContributorsOfCommits(Commit):
+    """
+    Class for New contributors of commits.
+
+    :param items: A list of dictionaries, each element a line from the
+        JSON file with Perceval data
+    :param date_range: A tuple which represents the start and end date of
+        interest
+    :param issrccode_obj: An object of IsSourceCode, used to determine
```
jgbarah: This parameter has a different name now, hasn't it?

implementations/code_df/new_contributors_of_commits.py
```diff
@@ -0,0 +1,147 @@
...
+                    (self.check_since <= self.df['created_date'])
+                    & (self.check_until >= self.df['created_date'])
+                    & (~self.df['author']
+                        .isin(
+                            list(committers_before_check_since)))
+                ]
+        count_new_committers = len(df_new_committers.unique())
+        return count_new_committers
+
+    def compute_timeseries(self, period='month'):
```
jgbarah: I still don't get why this is not done at the Metric, or maybe Commit, level. Could you please explain in which this code is different from other metrics, or other metrics in the Commit hierarchy?

jgbarah: I think there are (maybe) better ways of computing the time series. For example, you can make time the index of the dataframe, and then use `df.resample('M').sum()` (for months as time series units). Check some more background on this [here](https://chrisalbon.com/python/data_wrangling/pandas_time_series_basics/), for example.

jgbarah: For period, we could use Pandas "names" (eg, `"M"`). I see benefits in using complete English names (eg, `"month"`), but still...

Polaris000: Fine. I'll make the required changes.

Polaris000: I skimmed through the page and liked the idea a lot. I'll play around with it a bit and get back to you. Thanks!

Polaris000: I think this will really well. I'll integrate in into our project. Once again, thanks!

Polaris000: With the current implementation, I find it difficult to generalize a compute_timeseries method. There are several different components that change from implementation to implementation. 
- For example, for this metric I used a custom aggregation method, which I nested inside `compute_timeseries`. 
- In `open issue age`, I added another column to the dataframe for the age of that issue. I also used that column in the grouping. 

We could try the new method which may solve these problems.

Polaris000: That would also remove the need to have a conditional input check.

Polaris000:
Please have a look @jgbarah.

implementations/code_df/new_contributors_of_commits.py
jgbarah: Now, with the latest time_series method in your pull request, could you address these issues?

implementations/code_df/new_contributors_of_commits.py
```diff
@@ -0,0 +1,147 @@
...
+        """
```
jgbarah: Can we use the current `time_series` in your pull request, instead of this implementation?

Polaris000: Very sorry. I made the changes and made a commit for the notebook, forgetting to add this file. 
Please have a look now with the latest changes. Regarding your queries of why I am renaming functions, I have addressed them in the Code Changes Git pull request (#176)

Polaris000: Very sorry. I made the changes and made a commit for the notebook, forgetting to add the python  file to the commit. 
Please have a look now.

implementations/code_df/new_contributors_of_commits.py
```diff
@@ -0,0 +1,139 @@
...
+        :param period: A string which can be any one of the pandas time
+            series rules:
+            'W': week
+            'M': month
+            'D': day
+
+        :returns df: The final aggregated DataFrame
+        """
+
+        def __new_committers_count(series):
```
Polaris000: @jgbarah, according to our current implementation of `time_series()` and `_agg`, we would required implementing a separate aggregation method since we can't use the standard `count`, `mean`, etc. In this case, it is `__new_committers_count`. 

Would you prefer it being nested inside `_agg`  or as a separate but private/protected method?
Please let me know. The current implementation is, of course, a nested one.

Polaris000:
@jgbarah,I think it has been updated. 
- The `NewContributorsOfCommits` now has an overridden `_agg` method, defined (but raising`NotImplementedError`) in the Metric (root) class. 
- The `compute_timeseries` method has been renamed to `time_series` and is now defined in the Metric(root) class.
- Different parameters to the `compute` method( which allow excluding empty, merge commits, etc.) have been removed as well.

The thing that may seem different is the nested class. The problem here is that to find new committers in a given period (a week or a month), committers in that period must be checked with all committers from the start of the data to before the start of the current period to find out how many are first time committers.

To make this computation easier, I implemented a custom aggregation method which I have nested inside `_agg` so that the overall structure remains the same when compared to metrics which do not need a custom aggregation method. 

Hence, in the review **I** performed above, I asked whether you liked the nested aggregation method or did you want it outside as part of the `NewContributorsOfCommits` class, or even anywhere else.

jgbarah:
Thanks for the clarification on the nested class. Now I think I understand your code. However, two comments:

I think you have another, more simple, approach for the implementation: in `__init__`, in addition to the current df, you can build another one by grouping by author, and getting for each group the first date (which is when the author started to contribute). You can use the approach described at [this answer in SOF](https://stackoverflow.com/a/41525937), something like:

```
self.df = self.df.loc[df.groupby('author').date.idxmax()]
```

In other words, you change the dataframe for one which has a row per author, with the first time that person contributed as time field.

Then, in `_agg` you just count persons (rows)...

What do you think?

In any case, now I realize that we should also work on the definition of this metric, because it is not one of those that we're about to release. But that's not a big issue, so go ahead anyway.

Polaris000:
@jgbarah, I modified the code.
Please have a look. 

Though the code is now a lot cleaner, it strays from the other metrics, in that we completely modifying the dataframe. I am not sure how this becomes easier to someone trying to understand this code when compared to my implementation (adding a new column or a nested aggregation method), especially when the implementation you suggest changes the dataframe for the sake of easier calculation.

Could you please give me some pointers on this?

jgbarah:
> Though the code is now a lot cleaner, it strays from the other metrics, in that we completely modifying the dataframe. I am not sure how this becomes easier to someone trying to understand this code when compared to my implementation (adding a new column or a nested aggregation method), especially when the implementation you suggest changes the dataframe for the sake of easier calculation.
> 
> Could you please give me some pointers on this?

My idea is that every metric may need a different dataframe. For example, I think not all the fields in the current dataframe are needed, at least not for all metrics we're computing, But we can come back to that late.

With this implementation, you do have the "standard" dataframe, although you change it (by reducing it) in a further step...

Of course, we have to document all of this in detail, so that it doesn't confuse somebody ready the code.

Issue #175, opened 2019-06-18
[GSoC] Issue with `compute_timeseries` method in dataframe reference implementations 


Polaris000:
`compute_timeseries` method in the commit-related metric classes can be improved. The main issue is not exactly with the current implementation of `compute_timeseries`, rather, It is because the `self.since` and `self.until` parameters have to be set to particular values. This leads to them being assigned values in the `__init__` method of the classes.

Polaris000:
The current implemention of `compute_timeseries` of Code_Changes metric is like this:
- create a dataframe by grouping data based on `year`, followed by `month` of creation and then aggregating them with `count`.
- create a dataframe with rows representing every possible interval of time between the `since` and `until` parameters. For example, if the period is "month", then the second dataframe will have each year, followed by each month for that year between the `date_range` dates as rows.
- merge the two dataframes above

It is clear that creating the second dataframe will require a definite `since` and `until` date. The advantage of this method is that it allows for easier plotting of graphs, say, the number of commits per month. Only grouping the first dataframe would mean that months without commits will not be included at all, making a plot based on this dataframe skewed.

Polaris000:
Closing this since this was addressed in#176

PR #176, opened 2019-06-18
[GSoC] Reimplement compute_timeseries

Polaris000:
I have moved `self.since` and `self.until` population in case of `None` to `compute_timeseries`. This way, there is nothing out of place in `__init__` of `Commit` class in `commit.py`.

Also, all `__init__` methods in the `Algorithm` hierarchy now raise `NotImplementedError`.

### DIFF
--- a/implementations/code_df/code_changes_git.py
+++ b/implementations/code_df/code_changes_git.py
@@ -22,31 +22,46 @@ class CodeChangesGit(Commit):
 
         return len(self.df['hash'].unique())
 
-    @utils.rename_function('count')
-    def _aggregate(self, series):
+    def _agg(self, df, period):
         """
-        Perform an aggregation operation on a DataFrame or Series.
+        Perform an aggregation operation on a DataFrame or Series
+        to count the number of commits created in a every interval
+        of the period specified in the time_series method, like
+        'M', 'W',etc.
 
-        :param series: A pandas.Series object passed implicitly during
-        aggregation operations.
+        It simply counts the number of rows in the series, excluding
+        NaN values.
 
-        :returns: aggregation value, like mean, sum or count.
+        :param df: A pandas DataFrame on which to apply the aggregation
+
+        :param period: A string which can be any one of the pandas time
+        series rules:
+            'W': week
+            'M': month
+            'D': day
+
+        :returns count: the number of commits in the given series.
         """
 
-        return series.count()
+        df = df.resample(period)['category'].agg(['count'])
+        return df
 
 
 if __name__ == "__main__":
     date_since = datetime.strptime("2018-09-07", "%Y-%m-%d")
     items = utils.read_json_file('../git-commits.json')
+
     changes = CodeChangesGit(items, date_range=(date_since, None))
     print("Code_Changes, total:", changes.compute())
+
     changes = CodeChangesGit(items, date_range=(date_since, None),
                              is_code=[conditions.DirExclude(['tests']),
                                       conditions.PostfixExclude(['.md', 'COPYING'])])
     print("Code_Changes, excluding some files:", changes.compute())
+
     changes = CodeChangesGit(items, date_range=(date_since, None),
                              conds=[conditions.MasterInclude()])
     print("Code_Changes, only for master:", changes.compute())
+
     print("The number of commits created on a monthly basis is:")
-    print(changes.compute_time_series(period='M'))
+    print(changes.time_series(period='M'))
--- a/implementations/code_df/code_changes_git.py
+++ b/implementations/code_df/code_changes_git.py
@@ -1,145 +1,44 @@
 from datetime import datetime
 
-import pandas as pd
-
 import conditions
 import utils
 from commit import Commit
 
 
 class CodeChangesGit(Commit):
-    """Class for Code_Changes for Git repositories.
-
-    :param items: A list of dictionaries, each element a line from the
-        JSON file with Perceval data
-    :param date_range: A tuple which represents the start and end date of
-        interest
-    :param issrccode_obj: An object of IsSourceCode, used to determine
-        what comprises source code.
+    """
+    Class for Code_Changes for Git repositories.
     """
 
-    def compute(self, incl_empty=True, incl_merge=True, master_only=False):
-        """Count number of commits of different types, like including empty commits
-        or counting only those commits made on the master branch.
-
-        :param incl_empty: Include empty commits
-        :param incl_merge: Include merge commits
-        :param master_only: Include only commits made on master branch
-
-        :returns count: the number of commits satisfying the passed
-            conditions above
+    def compute(self):
         """
-        df = self.df
+        Count number of commits of different types, like including
+        empty commits or counting only those commits made on
+        the master branch.
 
-        if master_only:
-            count = self._count_master_only(incl_empty)
-
-        else:
-            if not incl_empty:
-                df = df[df['files_action'] != 0]
-            if not incl_merge:
-                df = df[~df['merge']]
-            count = df['hash'].nunique()
-
-        return count
-
-    def compute_timeseries(self, period='month'):
-        """
-        The metric value is computed for each fixed interval of time
-        from the "since" date to the "until" date arguments, specified
-        during object initiation.
-
-        The fixed time interval can be either a month or a week.
-
-        :param period: A string which can be either "month" or "week"
-
-        :returns dataframe: A DataFrame whose rows each represent an interval
-            of "period" and the count for that interval
+        :returns count: the number of commits satisfying the conditions passed
+            while instantiating CodeChangesGit.
         """
 
-        df = self.df
-        if period == 'month':
-            timeseries_series = df['created_date'] \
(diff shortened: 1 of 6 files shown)


Polaris000:
@jgbarah , please have a look whenever convenient :)

jgbarah:
Maybe you could rebase this (if you feel it is still convenient) after #177 was merged? Or maybe re-start from scratch, I m not sure what would be better...

Polaris000:
@jgbarah, following the patch you merged, in `conditions.py`, the class `Code` and its children all have a `pass` statement in their `__init__` methods ([here](https://github.com/chaoss/wg-evolution/blob/1a275d0c4688935309c9c88fd6e9c78bcc1dc2cc/implementations/code_df/conditions.py#L19)). But I think you wanted them to raise a `NotImplementedError` exception as you mentioned in previous reviews. Which one do you prefer?

I'll have a look as to what can be done with this pull request.

Polaris000:
Please have a look @jgbarah. This is a temporary solution to the problem of having `since` and `until` being populated in the `__init__` method of commit.py.

Polaris000:
Please have a look @jgbarah, @aswanipranjal.

Polaris000:
Sorry for the delay. I wasn't feeling to well the past couple of days. I'll add changes to the notebook if you approve of the changes.

Polaris000:
Closes #175.

implementations/code_df/metric.py
```diff
@@ -33,3 +34,45 @@ def _flatten(self, item):
...
         raise NotImplementedError
+
+    def _aggregate(self, series):
+        """
+        Aggregate the re-sampled DataFrame
+        """
+
+        raise NotImplementedError
+
+    def compute_time_series(self, cols=None, period='M'):
```
jgbarah: `time_series` could be a shorter, and still meaningful, name.

implementations/code_df/utils.py
```diff
@@ -66,3 +42,15 @@ def read_JSON_file(path):
 
             items.append(line)
     return items
+
+
+def rename_function(new_name):
```
jgbarah: I see you're using this, but I don't understand why we need it.

implementations/code_df/code_changes_git.py
```diff
@@ -1,152 +1,52 @@
...
-        timeseries_df = pd.DataFrame(timeseries_series)
-        timeseries_df.reset_index(inplace=True)
-        timeseries_df.columns = ['year', period, 'count']
-        merged_df = all_periods.merge(timeseries_df, how='outer').fillna(0)
-
-        dataframe = merged_df
-        return dataframe
-
-    def _count_master_only(self, incl_empty=True):
+    @utils.rename_function('count')
```
jgbarah: I don't get what are you doing here. Why renaming the function?

implementations/code_df/code_changes_git.py
```diff
@@ -1,152 +1,52 @@
...
-        Counts commits present only on the master branch.
+        Perform an aggregation operation on a DataFrame or Series.
 
-        :param incl_empty: exclude empty commits on the master branch
+        :param series: A pandas.Series object passed implicitly during
+        aggregation operations.
 
-        :returns commits_count: the number of commits created on the
-            master branch
+        :returns: aggregation value, like mean, sum or count.
```
jgbarah: But it seems this always counts... When is it computing other stuff, like mean or sum?

implementations/code_df/metric.py
Polaris000: Will do!

implementations/code_df/utils.py
Polaris000: It is an attempt at generalizing the `compute_time_series` method. The `agg` function by default, sets the name of the aggregation method used as the name of the resulting column post aggregation. 
For example, 
```python
df = df['abc'].resample(..).agg(function_obj)
df.columns = df.columns.droplevel(0)
```
would make the df look like this:
```
              'function_obj`
index
..                ...
..                ...
..                ...
..                ...
```

By changing the name of the method, we do not need the user to pass in a separate argument to `compute_time_series` to name the aggregated column.

What do you think? Are there better ways to do it / is this not required?

implementations/code_df/code_changes_git.py
Polaris000: Please check my comment above.

implementations/code_df/code_changes_git.py
Polaris000: Will make that more specific.

implementations/code_df/utils.py
Polaris000: Another approach could be this:
Use a method to add a column specifically for the `time_series` or `compute` method, like this one in Code_Changes_Lines:
```python
    def _add_lines_modified_cols(self, df):

        if len(df) == 0:
            raise ValueError("DataFrame empty. "
                             "Please check instantiation parameters")

        additions = list()
        deletions = list()
        for _, commit in df.iterrows():
            added_lines = 0
            removed_lines = 0
            for file in commit['files']:
                if 'action' in file:
                    try:
                        added_lines += int(file['added'])
                        removed_lines += int(file['removed'])
                    except:
                        pass

            additions.append(added_lines)
            deletions.append(removed_lines)

        df['additions'] = additions
        df['deletions'] = deletions

        return df
```

And then we would not require renaming the method. Of course, `agg` would still use the method name (`_aggregate`) to reference a partcular column. To fix that, we can modify the `time_series` method in `metric.py` to this:
```python
        df = self.df

        df = df.set_index('created_date')
        # No particular column to aggregate on:
        if cols is None:
            cols = ['category']

        df = df[cols].resample(period).agg([self._aggregate])
        df.columns = df.columns.droplevel(1)
```

The result of `time_series` of CodeChangesLines would be something like this:
```python
code_changes_lines_obj.time_series(cols=['additions', 'deletions']
```
```
            additions            deletions
index 
..             ...                  ...
..             ...                  ....
..             ...                  ...
..             ...                  ...
```
where `additions` and `deletions` are just names of columns in `df` on which we applied the aggregation.
I think this method is better for more complex metrics, which would required aggregations on several columns. What do you think @jgbarah @aswanipranjal?

Polaris000: The main problem is the naming of columns on performing aggregation.

jgbarah: i sort of understand now. I would do it this way (assuming for each specific metric, we're going to use a single aggregation method, which i guess is the case):

* Have a method, say `_agg`, which would get a df as argument, and would return another df aggregated as you want. This method would be abstract in Metric / Commit, and will be defined for each of the specific metric.

* At the level we define the method `time_series`, (either in class Metric or in class Commit) you have the generic code, that calls `_agg()` when appropriate.

Does this sound ok to you?

Polaris000: What I think you are saying is that `_agg` is not an aggregation method in the real sense. That is, it won't be passed to `Resampler.apply` or `Resampler.agg()`. Instead, it is just a regular method which takes a dataframe as input and returns whatever is required. Inside `_agg`, the resampling and aggregation happens.

The problem I think is that the aggregation requires passing a function object to the `agg` or `apply` method which is used after resampling:
```python3
df = df.resample('M').agg(my_agg_func)
```
And my_agg_func is passed a series: each column on which it is applied.
If we pass a dataframe to `_agg`, we would probably need another method to actually perform the aggregation (it will be passed to `Resampler.agg`).

The main motive behind the second method I suggested above is that for almost all of our metrics, we can add a few columns to `self.df` to allow for easier computation of metrics. Hence, we could have an overridden `_add_cols` method which adds columns to self.df. Also, whatever columns we added could be stored in a list.
Then, we could do this:
```python
        dict_ = {col: self._aggregate for col in self.added_cols}
        df = self.df
        df = df.set_index('created_date')
        df = df.resample(period).agg(dict_)
```
And self._aggregate would be: (for code_changes_lines)
```
    def _aggregate(self, series):
        total_modifications = series.sum()

        return total_modifications
```
What do you think?

Polaris000: Sorry if this sounds like a very obvious question.

jgbarah: I don't think we're understanding each other ;-) What I'm suggesting is rather simple. For the line that you have above:

```
df = df[cols].resample(period).agg([self._aggregate])
```

You would instead do something like:

```
df = self._agg(df)
```

And `_agg` would be like:

```
def _agg(self.df):

    return df.resample('M').count()
```

(assuming we're using months as the timeseries, and sum as the aggregating function. For a single metric, usually a single aggregation function is good enough. For example, .count() is the right one hen counting commits. Of course `_agg` for other metric could use a very different aggregation method.

Of course, you can also make this more general, defining a function which is the last one to be called on df.resample('M'), but maybe that's introducing an unneeded complexity. However, if you come with a clean solution, that could also be an option.

Is it more clear now?

Polaris000: Definitely more clear! Will get back to you tomorrow morning.

Polaris000:
Please have a look @jgbarah.

PR #177, opened 2019-06-18
[implementation] Restructuring reference implementations

jgbarah:
These commits restructure the reference implementations, adding "conditions" for checking if a file is considered as source code, or if a commit fulfills some condition (such as being part of the master branch).


implementations/code_df/commit.py
```diff
@@ -17,73 +19,87 @@ class Commit(Metric):
...
         who last falls inside the until range will be included.
 
-    :param issrccode_obj: A reference to IsSourceCode class.
+    :param is_code:  list of CodeCondition objects
         It is used to determine what comprises source code.
         """
 
-    def __init__(self, items, date_range=(None, None), issrccode_obj=None):
+    def __init__(self, items, date_range=(None, None), is_code=[conditions.Naive()],
+                 conds=[]):
```
Polaris000: @jgbarah, it would be better to set `conds` to `None`. [The problem.](https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument)

implementations/code_df/commit.py
```diff
@@ -1,14 +1,16 @@
 import pandas as pd
-from .metric import Metric
-from . import utils
+
+import conditions
+from metric import Metric
+import utils
```
Polaris000: There is a problem with importing modules this way. Since these modules are imported into the reference notebooks, which are in a different directory, using a relative import works. The new changes give a `ModuleNotFound` exception.
Python does not check for a modules location in the parent directory. As I am not a hundred percent sure of the fix, I'm open for inputs @valeriocos @aswanipranjal. :)

My current solution is to add `..` to sys.path and then change the intra package imports inside `code_df` (like in  commits.py) to relative imports. 

**notebook:**
```python
import sys
sys.path.append('..')
```
**commits.py:**
```python
from . import conditions
from . import utils
from .commit import Commit
```

Polaris000:
@jgbarah, is the goal of this pull request to let me know what changes should be made or will this pull request get merged? There are a few very minor problems in the doc strings of some methods.

implementations/code_df/conditions.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        Set the dataframe with commits to be analyzed for condition
+
+        This method also prepares the set of included commits
+        (those in master), so that the check can be done quickly later.
+
+        :param commits: commits (dataframe)
+        """
+
+        self.commits = commits
```
valeriocos: It could be useful to add some doc to explain how the collection of commits is done (starting from a the HEAD in master branch, the other commits are collected by traversing the parent links.

implementations/code_df/commit.py
Polaris000: @valeriocos , could you have a look at this? Thanks

aswanipranjal: Which directory are you running the script from, @Polaris000?

Polaris000: notebooks_df

aswanipranjal: for the notebooks, I think
```
import sys
sys.path.append('..')
```

would be the way to go

Polaris000: For that to work, commit.py should also be changed (as I mentioned above) so that it uses relative imports. Thanks @aswanipranjal!

aswanipranjal: correct, @Polaris000.

implementations/code_df/commit.py
jgbarah: Hmmmm. Thanks for the suggestion. However, I think that only applies if you change `conds` within the function, which is something yo don't do in this case, and would make little sense doing. Or maybe I'm wrong? Can you suggest some example where this would be a problem?

implementations/code_df/commit.py
jgbarah: Thanks a lot for all this discussion. If you don't mind, let's keep it as it is now, and let's figure out in a following pr how to deal with this in the context of Notebooks. When I proposed to have separate directories, I didn't notice this problem. Maybe we can revert to having just one directory, or find some other solution... Suggestions are welcome ;-)

jgbarah:
>  the goal of this pull request to let me know what changes should be made or will this pull request get merged? There are a few very minor problems in the doc strings of some methods.

I was intending to merge it, so that we have a more solid starting point. I think we can deal with docstrings and other details in further pull requests, if you all agree.

implementations/code_df/conditions.py
jgbarah: Good point. I will do.

foundjem:
LGTM @jgbarah you can merge it, and minor changes will happen later.

jgbarah:
OK, I'm merging it for now. Thanks a lot for all the comments. Let's continue from here. @Polaris000 (and everybody), in any case, please feel free to follow on the discussion in the items that were opened above.

implementations/code_df/commit.py
Polaris000: @jgbarah, I agree. `conds`, as well as `is_code` parameters are not getting modified. It is OK to leave it there. Or you could do something like this inside `__init__`:
```python
if cond is None:
    cond = list()
```

Polaris000:
A few suggestions: 
- We could also implement classes for Empty/Merge commits, just like MasterInclude and remove those parameters from `compute()`
- We can have a data directory which has three files: git_commit data, issue data and pull request data.
Let me know what you all think about these points. Thanks!

jgbarah:
> A few suggestions:
> 
>     * We could also implement classes for Empty/Merge commits, just like MasterInclude and remove those parameters from `compute()`

Yes, that was my idea. Those would be like a family of filters to the data, to be applied to "select" the items we're interested in.

>     * We can have a data directory which has three files: git_commit data, issue data and pull request data.

Good idea.

Polaris000:
Ok! Will implement that!

PR #180, opened 2019-06-21
[GSoC] Open issue age reference implementation and issue.py

Polaris000:
This pull request adds the reference implementation of `open issues age` metric. 
It also adds issue data fetched by Perceval, as a JSON file.


implementations/code_df/open_issue_age.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        super().__init__(data_list, date_range)
+
+        self.df = self._add_open_issue_age_col(self.df)
+
+    def compute(self):
+        """
+        Compute the average open issue age for all issues in the Perceval data.
+
+        :returns avg_open_issue_ages: the average age of open
```
aswanipranjal: I think the variable name should be singluar `avg_open_issue_age` since it's a single value and not, say, a list

implementations/code_df/open_issue_age.py
```diff
@@ -0,0 +1,138 @@
...
+        timeseries_df.columns = ["year", period, "open_issue_age"]
+        timeseries_df['open_issue_age'] = \
+            timeseries_df['open_issue_age'].round(decimals=2)
+
+        merged_df = all_periods.merge(timeseries_df, how='outer').fillna(0)
+
+        dataframe = merged_df
+        return dataframe
+
+    def _add_open_issue_age_col(self, df):
```
aswanipranjal: This could be renamed to `_calculate_open_issue_age` as it actually modifying the passed dataframe by calculating it and adding it to the df. What do you think @Polaris000?

implementations/code_df/issue.py
```diff
@@ -0,0 +1,55 @@
...
+        flat = {
+            'repo': item['origin'],
+            'hash': item['data']['id'],
+            'category': "issue",
+            'author': item['data']['user']['login'],
+            'created_date': creation_date,
+            'current_status': item['data']['state']
+        }
+
+        return [flat]
```
aswanipranjal: why do you need to return a list here?

implementations/code_df/open_issue_age.py
```diff
@@ -0,0 +1,138 @@
...
+            of "period" and the count for that interval.
+        """
+
+        if self.since is None:
+            self.since = utils.get_date(self.df, 'since')
+        if self.until is None:
+            self.until = utils.get_date(self.df, 'until')
+
+        df = self.df
+        if period == "month":
```
aswanipranjal: The code under `if` and `else` is almost the same. You could have a dictionary such as
```
period_frequency_map = {"month": "M", "week": "W", ....}
```

and use the `period` variable to determine what the frequency should be. What do you think @Polaris000?

implementations/code_df/open_issue_age.py
```diff
@@ -0,0 +1,138 @@
...
+
+        :raises ValueError: No analysis possible if no valid data in df.
+        """
+
+        if len(df) > 0:
+            # add open_issue_age column
+            df['open_issue_age'] = pd.np.nan
+
+            # replace NaN values for open issues by their "age"
+            issue_status_series = datetime.now() - df["created_date"][
```
aswanipranjal: could be renamed to `issue_statuses`

implementations/code_df/issue.py
```diff
@@ -0,0 +1,55 @@
+from metric import Metric
+import utils
+
+
+class Issue(Metric):
+
+    def __init__(self, items, date_range=(None, None)):
+        """
+        Initilizes self.df, the dataframe with one commit per row.
```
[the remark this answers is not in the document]
Polaris000: Oops. Will change that! Good catch.

implementations/code_df/open_issue_age.py
Polaris000: OK. I was going for "average" of all of open issue ages. But yours makes more sense. Will change it.

implementations/code_df/open_issue_age.py
Polaris000: No problem!

implementations/code_df/issue.py
Polaris000: This is part of Jesus' design. If you have a look at a part of the metric (root) class: 
```python
    def __init__(self, items):

        flat_items = []
        for item in items:
            flat_items.extend(self._flatten(item))
        self.df = pd.DataFrame(flat_items)
```

the flat_items list is extended by each flattened item. Hence, `_flatten` must pass a list so the `extend` can work

implementations/code_df/open_issue_age.py
Polaris000: Good point @aswanipranjal! @jgbarah probably noticed this issue and told me to pass the pandas timescale values like `'M', 'W',` etc directly to the parameter `period` instead "month", "week", etc. 

In any case, I am re-implementing `compute_time_series` using `DataFrame.resample(..)`

implementations/code_df/open_issue_age.py
Polaris000: Will do! I had used it to make it easier to identify it as being of type `pandas.Series`. The next iteration will have these changes. Thanks!

implementations/code_df/issue.py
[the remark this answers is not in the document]
Polaris000: Done!

implementations/code_df/issue.py
```diff
@@ -0,0 +1,55 @@
...
+
+        :param items: A list of dictionaries.
+            Each item is a Perceval dictionary, obtained from a JSON
+            file or from Perceval directly.
+
+        :param date_range: A tuple which represents the period of interest
+            It is of the form (since, until), where since and until are
+            strings. Either, or both can be None. If, for example, since
+            is None, that would mean that all commits from the first commit
+            to the commit who last falls inside the until range will be
```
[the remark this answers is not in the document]
Polaris000: done

implementations/code_df/open_issue_age.py
Polaris000: Done

Polaris000:
Please have a look @aswanipranjal @jgbarah

implementations/code_df/issue.py
jgbarah: The idea is that in some cases, _flat could return more than one row. For example, if you need to "convert" a commit in information for each file touched by that commit.

Polaris000:
Please have a look @jgbarah. Once this is merged, I'll change the naming for all files in a single pull request, hopefully tomorrow.

implementations/code_df/issue.py
```diff
@@ -0,0 +1,61 @@
...
+            return []
+
+        if self.until and (self.until < creation_date):
+            return []
+    
+        # all pull requests are considered to be issues by default.
+        # this check removes pull requests from the
+        # analysis.
+        if 'pull_request' in item['data']:
+            return []
```
Polaris000: This condition is to avoid analyzing pull requests in the data. All pull requests are treated as issues and hence the issues data also includes pull requests.

implementations/code_df/open_issue_age.py
```diff
@@ -0,0 +1,108 @@
...
+            'category': "issue",
+            'author': item['data']['user']['login'],
+            'created_date': creation_date,
+            'current_status': item['data']['state']
+        }
+
+        if flat['current_status'] == 'open':
+            flat['open_issue_age'] = \
+               (datetime.now() - flat['created_date']).days
+
```
Polaris000: The open issue age field is calculated by simply taking the days value from the timedelta object. The `if` condition makes it so that the final `df` has only `open` issues with valid values for the `open_issue_age` column. Do you agree with this?

implementations/code_df/issue.py
jgbarah: I think we can assume all items are pull requests. In other words, we can assume that the filtering is done before the data arrives into this class. So instead of including this check, let's include a comment in the description of the class stating that it should be initialized with a list of pull requests.

implementations/code_df/open_issue_age.py
jgbarah: I think open issue is defined only for issues that were closed... However, since that metric is still not fixed in stone, proceed as you want, and we check that later.

jgbarah:
Please, have a look at the comments, mainly at the one on assuming all items are pull requests.

implementations/code_df/issue.py
Polaris000: > I think we can assume all items are pull requests.

@jgbarah, this is an issue heirarchy metric and hence, should be used on issue data, not pull requests. I think you meant to say this.

> So instead of including this check, let's include a comment in the description of the class stating that it should be initialized with a list of pull requests.

I think you mean to say, "... that it should be initialized with a list of issues." I think this will be difficult to do since the only differences between regular pull requests and those pull requests in the issue data fetched by Perceval are a few fields. 

Please just confirm what you meant and I'll remove the condition to **remove pull requests and not issues** and add a comment as required. Thanks!

implementations/code_df/open_issue_age.py
Polaris000: Ok. If the final definition deviates from this, I'll update the implementation with another patch.

Polaris000:
@jgbarah. I have responded to your comments above. Please have a look whenever possible. Thanks for the review!

implementations/code_df/issue.py
jgbarah: Oooops. Yes. Let's assume Perceval items are filtered before they are fed to this class, and all of them are issues. So, we wouldn't need to filter them when initlalizing the object. Sorry for the noise.

The code for filtering in only issues is simple:

```
issues = [item if pull_request not in item['data'] for item in items]
```
(code not tested, but should be like that)

implementations/code_df/open_issue_age.py
jgbarah: OK.

jgbarah:
Done. Please proceed as commented, and we can merge this one. Thanks!

Polaris000:
Have a look @jgbarah.

PR #190, opened 2019-06-23
[GSoC] Reference implementation for Code_Changes_Lines

Polaris000:
This pull request adds the reference implementation (notebook + python script) for the metric Code_Changes_Lines.


Polaris000:
Please have a look @jgbarah.

implementations/code_df/code_changes_lines.py
```diff
@@ -0,0 +1,128 @@
...
+    :param is_code:  list of CodeCondition objects
+        It is used to determine what comprises source code.
+
+    :param conds: A list of Commit objects. It is used to
+        exclude or include different kinds of commits, like
+        empty or merge commits, or commits made on the master
+        branch.
+    """
+
+    def __init__(self, items, date_range=(None, None),
```
jgbarah: Isn't __init__ in the Commit class now?

implementations/code_df/code_changes_lines.py
```diff
@@ -0,0 +1,128 @@
...
+
+        :returns modified: A tuple representing the number of lines
+            modified. modified -> (additions, deletions)
+        """
+
+        df = self.df
+        additions = df['additions'].sum()
+        deletions = df['deletions'].sum()
+
+        modified = (additions, deletions)
```
jgbarah: The definiton for "modified" is "added" `"deleted", see description of the metric. Or am I wrong?

implementations/code_df/code_changes_lines.py
```diff
@@ -0,0 +1,128 @@
...
+            series rules:
+            'W': week
+            'M': month
+            'D': day
+
+        :returns df: The aggregated dataframe, where aggregations have
+            been perform on "additions" and "deletions" columns
+        """
+
+        df = df.resample(period).agg({"additions": 'sum', "deletions": 'sum'})
```
jgbarah: Not sure this is working: changed = added + deleted...

implementations/code_df/code_changes_lines.py
Polaris000: So the problem here is that unlike the other metrics we have implemented, this one cannot be directly computed by `sum` or `count` operations. I called the `__init__` method here so that I can add columns for modifications to the dataframe.
We can go about this in two ways:
- **Not add a column**  
Here would have to iterate over each file in every commit and add the number of lines modified to a "count" variable.
That would mean be a little messy, especially when it comes to aggregating over it.

- **Adding a column** 
This would allow for easier computation of the metric, and it is easier to understand for someone studying the code. The only problem, as of now, is having to add the column in the `__init__ ` method, which I have done here.

implementations/code_df/code_changes_lines.py
Polaris000: Thanks for the clarification! Will change the implementation.

implementations/code_df/code_changes_lines.py
Polaris000: Will change it

Polaris000:
@jgbarah, I have updated the implementation according to the correct definition of Code_Changes_Lines. If you do not prefer adding another column for modifications in the `__init__` method, we could have a function to compute the total line changes given a DataFrame and use it as the `compute` method, while for the `_agg` method, we could define a custom aggregation method like I did for `New_Contributors_to_Commits` in #172.

It seems to me that we have a choice between the following:
- Adding a column to the DataFrame and then using a standard aggregation method (sum, count, etc..)
- Not adding a column but then using a custom aggregation method (like New_Contributors_to_Commits).

If it is a bit confusing, the `__init__` method has been overridden in the metric class here so that I could do the first option mentioned above, which is to add a column.

What do you think?

implementations/code_df/code_changes_lines.py
jgbarah: Now I understand. Thanks for the clarification.

Consider the following approach: you can redefine _flatten, and for each commit, you can compute the number of lines added and removed in all files, and get from them the total number of lines modified, and build that directly into the flattened dictionary (a new column, "modified", with the number of lines modified).

Then, you can write a `_agg` file that uses `sum()` for the aggregation.

I think that would be cleaner... (and maybe  reason for having _agg at the level of the specific metric, instead than at the Commit class level, but we will see...)

implementations/code_df/code_changes_lines.py
jgbarah: Thanks!

jgbarah:
Thanks for your suggestions. Please, consider implementing an specific `_agg` method, as I commented above.

implementations/code_df/code_changes_lines.py
Polaris000: That would mean modifying the `_flatten` method in the `Commit` class. Is it a good idea to modify a parent class?

Polaris000:
Please have a look @jgbarah .

implementations/code_df/code_changes_lines.py
jgbarah: Let's see when it is implemented, but methods in a class hierarchy are written to be overridden. That's the idea of subclassing ;-)

Polaris000: Yes @jgbarah, my only problem is that this won't be the case for all metrics, meaning that all implementation would not have a similar structure (not all of them would have an overridden _flatten method).

Anyways, thanks for approving it.

PR #194, opened 2019-07-01
[GSoC] Add reference implementation for Review Accepted

Polaris000:
This pull request adds the reference implementation python script for Reviews Accepted.
The notebook will be added in a short while.

### DIFF
--- a/implementations/code_df/reviews_accepted.py
+++ b/implementations/code_df/reviews_accepted.py
@@ -1,0 +1,65 @@
+from datetime import datetime
+
+import utils
+from pullrequest import PullRequest
+
+
+class ReviewsAccepted(PullRequest):
+    """
+    Class for Reviews Accepted
+    """
+
+    def compute(self):
+        """
+        Compute the total number of reviews which were accepted, from the Perceval data.
+
+        :returns count: The total number of reviews accepted
+        """
+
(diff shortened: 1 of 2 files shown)


implementations/code_df/reviews_accepted.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+        Compute the total number of reviews which were accepted, from the Perceval data.
+
+        :returns count: The total number of reviews accepted
+        """
+
+        count = len(self.df['hash'][self.df['merged']].unique())
+        return count
+
+    def _agg(self, df, period):
```
jgbarah: Any reason for this not using self.df, instead of receiving df as paramenter? I know we have that in the other metric implentations, but I only realized now..

Polaris000: The `time_series` method in the Metric (root) class goes like this:
```python
    def time_series(self, period='M'):
        df = self.df

        df = df.set_index('created_date')
        df = self._agg(df, period)

        return df
```
In the implementation we have, if we use `self.df` directly in `_agg`, there is not much point in having the `time_series` method, since we would now be creating and modifying `df` inside `_agg`.

I'll open an issue and try to come up with a solution which does not require passing a dataframe as a parameter.
Thanks for pointing this out!

PR #200, opened 2019-07-03
[GSoC] Add  tests for commit-based DataFrame reference implementations

Polaris000:
This request is for adding tests for the DataFrame implementation of metrics.


Polaris000:
@jgbarah, I am not completely done with these and hence, am opening a draft request. 
Some doubts I had:
1. Since the Metric (root) class will never be instantiated, should I even test it? Any methods it has will either be inherited or overridden in child classes, so these methods can be tested in the child classes.
What do you think?  

2. Where should I test the `time_series` method?   
It is defined in the Metric (root) class but is used (following instantiation) in individual metric classes.   

3. Following 1., I am testing the method by using a temporary child class of Metric (root). In this class the `_agg` method will be trivially implemented.   

4. Following 2., since the `_flatten` method in Metric(root) class is not implemented, I would have to define it in the temporary class I create in point 2. Is there a better way to do this or should I just copy the `_flatten` method we have in `Commit` class?  

5. While testing the `_filterout` method in `Commit` class, I have again made use of a temporary class, since this method is called while instantiating the `Commit` class. Let me know if this is OK.

6. You will find several tests not implemented (just have a pass statement). I have not done those for now. Most of them have a similar test which I have implemented. For example, `test__filterout()` has been implemented while `test__filteroutMasterInclude()` has not. 

7. Tests not yet implemented:
- testing conditions.py
- testing the commit heirarchy (metric tests)

Please have a look and let me know if anything has to be changed / any other tips and suggestions.

jgbarah:
>     1. Since the Metric (root) class will never be instantiated, should I even test it? Any methods it has will either be inherited or overridden in child classes, so these methods can be tested in the child classes.
>        What do you think?

I think we don't need to test them, at least for now.
 
>     2. Where should I test the `time_series` method?
>        It is defined in the Metric (root) class but is used (following instantiation) in individual metric classes.

For now, in the specific metrics when you are testing it with real dfs, but it woudl be nice if you can fidn a way of testing the root implementation by itself (maybe setting a fake self.df to do the testing).

>     3. Following 1., I am testing the method by using a temporary child class of Metric (root). In this class the `_agg` method will be trivially implemented.

We can leave it as such for now, and decide later if to change it.

>     4. Following 2., since the `_flatten` method in Metric(root) class is not implemented, I would have to define it in the temporary class I create in point 2. Is there a better way to do this or should I just copy the `_flatten` method we have in `Commit` class?

You can provide an specific _flatten method. If needed, subclass for that.
 
>     5. While testing the `_filterout` method in `Commit` class, I have again made use of a temporary class, since this method is called while instantiating the `Commit` class. Let me know if this is OK.

OK, as above.
 
>     6. You will find several tests not implemented (just have a pass statement). I have not done those for now. Most of them have a similar test which I have implemented. For example, `test__filterout()` has been implemented while `test__filteroutMasterInclude()` has not.

As i said int he review, let's finish this, and you can work on those in a separate pr.
 
>     7. Tests not yet implemented:

Same as above.

Polaris000:
Have a look @jgbarah

Polaris000:
@jgbarah, I have added docstrings. Please have a look at them and let me know if further changes have to be made. Thanks.

PR #202, opened 2019-07-08
[GSoC] Add reference implementation for Issue Response Time

Polaris000:
This pull request adds the reference implementation for Issue Response Time along with tests.

### DIFF
--- a/implementations/code_df/issue_response_time_github.py
+++ b/implementations/code_df/issue_response_time_github.py
@@ -1,0 +1,114 @@
+from datetime import datetime
+import numpy as np
+
+import utils
+from issue import Issue
+
+
+class IssueResponseTimeGithub(Issue):
+    """
+    Class for Issue Response Time
+    """
+
+    def _flatten(self, item):
+        """
+        Flatten a raw issue fetched by Perceval into a flat dictionary.
+
+        A list with a single flat directory will be returned.
+        That dictionary will have the elements we need for computing metrics.
+        The list may be empty, if for some reason the issue should not
+        be considered.
+
+        :param item: raw item fetched by Perceval (dictionary)
+        :returns:   list of a single flat dictionary
+        """
+
+        creation_date = utils.str_to_date(item['data']['created_at'])
+        if self.since and (self.since > creation_date):
+            return []
+
+        if self.until and (self.until < creation_date):
+            return []
+
+        # all pull requests are considered to be issues by default.
+        # this check removes pull requests from the
+        # analysis.
+        if 'pull_request' in item['data']:
+            return []
+
+        flat = {
+            'repo': item['origin'],
+            'hash': item['data']['id'],
+            'category': "issue",
+            'author': item['data']['user']['login'],
+            'created_date': creation_date,
+            'current_status': item['data']['state']
+        }
+
+        member_comments = [comment for comment in item['data']['comments_data']
+                           if comment['author_association'] == 'MEMBER']
+
+        flat['issue_resp_time'] = np.NaN if len(member_comments) == 0 \
+            else (datetime.now()
+                  - utils.str_to_date(
+                    min(comment['created_at']
+                        for comment in member_comments))) \
+            .days
+
+        return [flat]
+
+    def compute(self):
+        """
+        Compute the average issue response time for all issues
+        in the Perceval data.
+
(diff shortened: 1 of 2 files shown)


Polaris000:
@jgbarah, please have a look at the implementation. I'll add the notebook and tests tomorrow. Let me know if you think it is ok.

implementations/code_df/issue_response_time_github.py
```diff
@@ -0,0 +1,114 @@
...
+            return []
+
+        if self.until and (self.until < creation_date):
+            return []
+
+        # all pull requests are considered to be issues by default.
+        # this check removes pull requests from the
+        # analysis.
+        if 'pull_request' in item['data']:
+            return []
```
Polaris000: Here I am removing pull requests from the issues data (since GitHub treats all pull requests as issues too). Hope this is fine. (this check will of course be in the `Issue_Github` category class (in `_flatten`) and is not specific to this metric.)

implementations/code_df/issue_response_time_github.py
```diff
@@ -0,0 +1,114 @@
...
+            'created_date': creation_date,
+            'current_status': item['data']['state']
+        }
+
+        member_comments = [comment for comment in item['data']['comments_data']
+                           if comment['author_association'] == 'MEMBER']
+
+        flat['issue_resp_time'] = np.NaN if len(member_comments) == 0 \
+            else (datetime.now()
+                  - utils.str_to_date(
```
Polaris000: To find the first comment by a maintainer, I used the following logic. Please have a look and let me know if this is easy to understand / requires other changes. What I am doing here is this:
- Pickout the comments by maintainers and store in `member_comments`.
- Add the issue response time to `flat` by calculated the number of days between today and the first comment by a maintainer (using the `min`).

implementations/code_df/issue_response_time_github.py
```diff
@@ -0,0 +1,114 @@
...
+            'D': day
+
+        :returns df: The aggregated dataframe, where aggregations have are the
+        so called items
+            been performed on the "issue_resp_time" column
+        """
+
+        df = df.resample(period).agg({'issue_resp_time': 'mean'})
+        df['issue_resp_time'] = df['issue_resp_time'].fillna(0)
+
```
Polaris000: The `fillna(0)` is required to make it easier to plot the dataframe in notebooks. I could just as well remove this line and in the notebook, just before plotting, I could run this line. What do you prefer?

jgbarah:
Before entering into details, I think we should focus on metrics that already done, at least done enough to be relases. Those are the ones [listed here](https://github.com/chaoss/wg-evolution/blob/master/focus_areas/code_development.md#summary-of-metrics).

Until we're done with those, let's keep this one in the fridge. I'll try to work on its definition meanwhile.

implementations/code_df/issue_response_time_github.py
jgbarah: As in other cases, i think it is easier for this reference implementation to just assume that the class is fed only with prs. It is easy to do that before instantiating the class. So, please remove that check...

implementations/code_df/issue_response_time_github.py
jgbarah: I don't understand this rationale. Time should be from the moment the issue is opened, to the moment it receives the first reponse (not from the first comment to today). If there was still no comment, we should ignore the issue, since it is not correct to say it was responded today...

Are we in sync?

implementations/code_df/issue_response_time_github.py
jgbarah: I think it is better if you do it when plotting. Otherwise, you're assuming that this is being called before plotting, which maybe is not the case...

Polaris000:
Hi @sgoggins. Thank you for your interest.
 I'll try my best to do this over the weekend.  If not, I'll work on it in the first week(s) of October, since I'm a little busy, right now. Sorry :(

jgbarah:
@Polaris000, if you don't find the time, don't worry. This metric is not fully defined anyway, so we could just reopen a pr later, when that's done.

jgbarah:
Since I see no movement, I´m closing this for now. @Polaris000, please feel free to open a new pr when you have time.

PR #203, opened 2019-07-14
[GSoC] Add non-pandas reference implementation for CodeChanges metric

### LINKED ISSUE #156: [GSoC] Create reference implementations without using pandas
This discussion was started in the latest weekly IRC chat with my mentors (the log can be found [here](https://github.com/Polaris000/GSoC_19_Perceval_Implementations/blob/master/IRC_chats/may_31_comm_bonding.txt)) but not finished. The idea is that reference implementations for metrics do not require the use of the Pandas library, meaning that each metric can be implemented purely with the use of native Python data structures. 

This issue is for discussing this idea. Everyone is welcome to share their views!

Polaris000:
This pull request adds the root class, hierarchy classes and utility classes along with the reference implementation for CodeChanges metric.

- Closes #156

### DIFF
--- a/implementations/code/code_changes_git.py
+++ b/implementations/code/code_changes_git.py
@@ -1,0 +1,42 @@
+from datetime import datetime
+
+import conditions
+import utils
+from commit_git import CommitGit
+
+
+class CodeChangesGit(CommitGit):
+    """
+    Class for Code_Changes for Git repositories.
+    """
+
+    def compute(self):
+        """
+        Count number of commits of different types, like including
+        empty commits or counting only those commits made on
+        the master branch.
+
+        :returns count: the number of commits satisfying the conditions passed
+            while instantiating CodeChangesGit.
+        """
+
+        commit_hashes = {item['hash'] for item in self.items}
+        return len(commit_hashes)
+
+
+if __name__ == "__main__":
+    date_since = datetime.strptime("2018-09-07", "%Y-%m-%d")
+    items = utils.read_json_file('../git-commits.json')
+
+    changes = CodeChangesGit(items, date_range=(date_since, None))
+    print("Code_Changes, total:", changes.compute())
+
+    changes = CodeChangesGit(items, date_range=(date_since, None),
+                             is_code=[conditions.DirExclude(['tests']),
+                                      conditions.PostfixExclude(
+                                        ['.md', 'COPYING'])])
+    print("Code_Changes, excluding some files:", changes.compute())
+
+    changes = CodeChangesGit(items, date_range=(date_since, None),
+                             conds=[conditions.MasterInclude()])
+    print("Code_Changes, only for master:", changes.compute())
--- a/implementations/code/commit_git.py
+++ b/implementations/code/commit_git.py
@@ -1,0 +1,102 @@
+import conditions
+from metric import Metric
+import utils
+
+
+class CommitGit(Metric):
+    """
+    Initializes self.items, a list with items(dictionary) as
+    elements.
+
+    :param items: A list of dictionaries.
+        Each item is a Perceval dictionary, obtained from a JSON
+        file or from Perceval directly.
+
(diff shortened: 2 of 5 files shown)


implementations/code/code_changes_git.py
```diff
@@ -0,0 +1,78 @@
...
+        timeseries = dict()
+        items_groupby = groupby(
+            items, key=lambda x: (x['created_date'].year,
+                                  x['created_date'].month))
+
+        for period, group in items_groupby:
+            group = list(group)
+
+            timeseries[period] = len(group)
+
```
Polaris000: I'm having a little trouble generating a timeseries similar to the result of the `resample()` method in the `pandas` implementation.
- For example, `df.resample('M').count()` performs the count aggregation operation on the groups but also returns a value of 0 for those groups with no elements. It is hard to do this with the `itertools` I am using.
- `df.resample()` allows for easily passing required timeperiods like `M` or `W`. Though I can use `itertool` to groupby month like the current implementation I have done, I'll have to come up with a different way to do it, so that it is easier for a period of a `week` or `day`.

Polaris000:
closes #156

implementations/code/code_changes_git.py
jgbarah: First of all, we don't need to do all the fancy stuff that Pandas resample does. We can stick to the basics, or even forget about the timeseries stuff, since we really don't need it for showing the definition of the metric.

I suggest that we start without timeseries, and then in a separate pull request, we discuss about the best way of integrating it. If we find a reasonable way, that's it. If not, we can live without that functionality for this "branch" of the reference implementation.

In any case, for having months with zero items, you can just iterate after you have the dictionary with all months with non-zero items, filling in the gaps. You can do that at the same time you convert the data to a list (which maybe is the best representation for the resulting time serie). That is, you iterate from first month to last month, and if there is nothing in the dictonary for that month, the result is zero. If there is, the result is that one.

You can also try to do that when you build the dictionary, using itertools to iterate on the list of months, but that's a bit complex and I m not sure it's doable without coding it myself.

implementations/code/code_changes_git.py
```diff
@@ -0,0 +1,78 @@
+from datetime import datetime
+from itertools import groupby
+import conditions
+import utils
+from commit_git import CommitGit
```
jgbarah: Please, leave a blank line. See [coding standard on import](https://www.python.org/dev/peps/pep-0008/#imports)

implementations/code/code_changes_git.py
```diff
@@ -0,0 +1,78 @@
...
+        """
+        Count number of commits of different types, like including
+        empty commits or counting only those commits made on
+        the master branch.
+
+        :returns count: the number of commits satisfying the conditions passed
+            while instantiating CodeChangesGit.
+        """
+
+        commit_hashes = [item['hash'] for item in self.items]
```
jgbarah: To avoid this, you can either make commit_hashes directly a set, or make it a dictionary, and then just count the dictionary (adding an existing key to a dictionary does not add a new element)

implementations/code/code_changes_git.py
Polaris000: > I suggest that we start without timeseries, and then in a separate pull request, we discuss about the best way of integrating it. If we find a reasonable way, that's it. If not, we can live without that functionality for this "branch" of the reference implementation.

Sure. I am removing the functionality for now.
Thanks for the other suggestions as well!

implementations/code/code_changes_git.py
Polaris000: Oops! Overlooked that. Made the change.

implementations/code/code_changes_git.py
Polaris000: Ok. Used a set comprehension to directly make `commit_hashes` a set.

Polaris000:
Please have a look @jgbarah.

PR #207, opened 2019-07-16
[GSoC] Add execution script

Polaris000:
This is an execution script to run metrics on all data. Currently, it is under development.


implementations/exec.py
```diff
@@ -0,0 +1,59 @@
...
+Git_metrics = [code_changes_git, code_changes_lines]
+GitHub_metrics = [reviews, ]
+
+
+def write_results(file=None, Git_results=None, GitHub_results=None):
+    print(Git_results)
+    print(GitHub_results)
+
+
+def test_git_metrics(items_file, date_range, is_code=None, conds=None):
```
valeriocos: Why the method is called `test..`?

implementations/exec.py
```diff
@@ -0,0 +1,59 @@
...
+    Git_results = {}
+
+    for metric in Git_metrics:
+        metric_obj = metric(items, date_range)
+        Git_results[str(metric)] = metric_obj.compute()
+
+    return Git_results
+
+
+def test_github_metrics(items_file, date_range):
```
valeriocos: This method looks really similar to `test_git_metrics`. Probably, it could be generalized. For instance, instead of having a list of metrics for each data source (see lines 19-20), we could have a dict that relates data sources to metrics like this one:
```
{
  'git': [code_changes_git, code_changes_lines],
  'github': [reviews]
}
```

The map will be read in the `main`, for each key (e.g., `git`, `github`), the corresponding Perceval backend will be initialized and executed. The obtained items, together with the values of the dict (e.g., `[code_changes_git, code_changes_lines]`) will be passed to the generalized version of `test_git_metrics`, and the results returned. WDYT?

implementations/exec.py
```diff
@@ -0,0 +1,59 @@
...
+
+def main(use_pandas=False, include_timeseries=False, git=True, github=False, ):
+
+    date_since = datetime.strptime("2018-09-07", "%Y-%m-%d")
+    Git_results = test_git_metrics('git-commits.json', (date_since, None))
+    write_results(Git_results=Git_results)
+
+
+if __name__ == '__main__':
+    main()
```
valeriocos: It would be great to add before `main()` the possibility to parse arguments from command line. In case you need an example to use `argparse, you can check: https://gist.github.com/valeriocos/f75e8081c9ca121851a2fb7def93d43b#file-scava_stats-py-L52

implementations/exec.py
valeriocos: It would be better if the method receives a list/generator of files instead of using `utils.read_json_file(items_file)`, thus the method will be really focused on calculating the metrics. The JSON files could be read in `main`, and this would help replacing the JSON files with a call to Perceval. WDYT?

Polaris000: Good idea! Will change that.

implementations/exec.py
Polaris000: Thanks valeriocos. Yes, the final version will use command line args.

Polaris000:
@valeriocos, I tried to implement several of your suggestions. Though this is in no way completely finished (have to add proper doc strings / refactor code a little / add proper defaults and help to argparse arguments), if possible, could you have a look at the script?
The script generates a pdf, which can be downloaded from the link below (the output pdf can be formatted later via the FPDF library)
[xyz.pdf](https://github.com/chaoss/wg-evolution/files/3452943/xyz.pdf)

A few queries:
- would it be a good idea to add a `__str__` method to each of the metric classes? It becomes easier to display metric names (especially when using generalized code).
- I think it would be good to add a `plot_chart` method to each metric class. Since each metric has different axes names to plot, it is difficult to generalize this. What do you think?
- > This method looks really similar to test_git_metrics. Probably, it could be generalized. For instance, instead of having a list of metrics for each data source (see lines 19-20), we could have a dict that relates data sources to metrics like this one:
```python
{
  'git': [code_changes_git, code_changes_lines],
  'github': [reviews]
}
```
I have structured to metrics like this:
```python
{
    `commit`: [<commit metric class 1>, ...],
    `issue`:[<issue metric class 1>, ...],
    `pull_request`:...
}
```
And the final result like this:
```python
{
    `commit`: [
        {
             'metric': metric_obj,
             'val': value of compute() method for that metric,
             'df': dataframe returned from timeseries method for that metric
         }, 
       {<metric 2>},
       {<metric 3>}...
    ],
    `issue`:[<same as above, but for issue classes>],
    `pull_request`:...
}
```

implementations/exec.py
Polaris000: Please see my comment after this review.

Polaris000:
I have added support for images and json and pdf support was already there. I could n't find a good way to add markdown support, but am still looking. I also made it executable from the command line, so you can do :
```bash
$ analyze -r chaoss/wg-evolution -o pdf images ....
```
Please have a look.

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+                        default=[Naive()],
+                        help="is code conds")
+
+    parser.add_argument("-t", "--time-series",
+                        action='store_true',
+                        help="To get plots of time-series'")
+
+    parser.add_argument("-pe", "--period",
+                        default='M',
+                        help="period for time-series: 'M', 'W', etc.")
```
valeriocos: If the periods are limited, it's probably better to add all of them. WDYT?

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+                    )
+
+    generate_output(results, args.output_formats)
+
+
+if __name__ == '__main__':
+    try:
+        main()
+    except KeyboardInterrupt:
+        sys.exit(0)
```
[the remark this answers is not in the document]
Polaris000: Good idea! will do.

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+
+            ]
+
+ISSUE_METRICS = []
+
+PULLREQUEST_METRICS = []
+
+METRICS = \
+    {
+        "commit": COMMIT_METRICS,
```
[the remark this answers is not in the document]
Polaris000: Sure.

bin/analyze
```diff
@@ -0,0 +1,283 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
Polaris000: OK.

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+    parser.add_argument("-c", "--conditions",
+                        default=[],
+                        nargs='*',
+                        type=list,
+                        help="Commit conds")
+
+    parser.add_argument("-x", "--is-code",
+                        type=list,
+                        default=[Naive()],
+                        help="is code conds")
```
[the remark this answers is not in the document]
Polaris000: `is_code` takes in different `Code` child objects to decide which commits to include in the analysis. 
For now, the following algorithms are supported:
-  Naive
- DirExclude
- PostfixExclude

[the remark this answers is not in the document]
valeriocos: Thank you for the explanation, could it make sense to have them as `choices`?

valeriocos:
Please @Polaris000 ping me when the PR is ready to review, thanks

Polaris000:
@valeriocos, I'll submit changes as soon as possible. I was working on another patch, but will switch to this now. Sorry to keep you waiting :(

valeriocos:
no worries @Polaris000 :) take your time, thanks

Polaris000:
If it is possible, you could have a look at my response to your review for #237

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+    code_changes_git,
+    code_changes_lines,
+    )
+from implementations.code_df.utils import str_to_date
+
+GITHUB_URI = "http://github.com/"
+COMMIT_METRICS = \
+            [
+                code_changes_git.CodeChangesGit,
+                code_changes_lines.CodeChangesLines,
```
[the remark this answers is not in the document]
Polaris000: done!

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+import argparse
+import os
+import sys
+
+from fpdf import FPDF
+import json
+
+from implementations.code_df.conditions import Naive
+
+from implementations.code_df import (
```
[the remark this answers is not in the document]
Polaris000: That's strange. I am getting the error too now, but I wouldn't have pushed it if it wasn't working.
Would it matter where you ran analyze?
Making the required changes.

bin/analyze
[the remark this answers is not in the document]
Polaris000: Fixed the datetime issue too!

bin/analyze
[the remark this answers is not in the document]
Polaris000: Yes, I'll probably have to add it to every module.

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+        "pull_request": PULLREQUEST_METRICS
+    }
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Analyze script argument parser"
+    )
+
+    parser.add_argument("-a", "--api-token",
```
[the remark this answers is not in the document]
Polaris000: Done!

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+                        help="GitHub API token", default=None)
+
+    parser.add_argument("-r", "--repo",
+                        required=True,
+                        help="GitHub repository, as 'owner/repo'")
+    parser.add_argument("-s", "--since",
+                        default=None,
+                        help="From date")
```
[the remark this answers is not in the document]
Polaris000: Done!

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+    parser.add_argument("-u", "--until",
+                        default=None,
+                        help="To date")
```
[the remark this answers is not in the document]
Polaris000: Done!

bin/analyze
```diff
@@ -0,0 +1,283 @@
...
+                        default=['commit'],
+                        nargs='+',
+                        choices=['commit', 'issue', 'pull_request'],
+                        help="commit, issue, pull_request (any combination)")
+
+    parser.add_argument("-c", "--conditions",
+                        default=[],
+                        nargs='*',
+                        type=list,
+                        help="Commit conds")
```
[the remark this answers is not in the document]
Polaris000: Done!

bin/analyze
Polaris000: https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases.
I think they are too many to list in the help section .... WDYT?

bin/analyze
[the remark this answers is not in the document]
Polaris000: I think choices allows for only one option. A user can pass in any combination of the above.
```python
    parser.add_argument("-i", "--is-code",
                        default=['Naive'],
                        nargs='+',
                        choices=['Naive', 'PostfixExclude', 'DirExclude'],
                        help="Definition of Source Code."
                             "Possible options: %(choices)s (any combination)")
```

I have implemented the following (coming iteration). WDYT?

bin/analyze
valeriocos: yes, I agree with you!

bin/analyze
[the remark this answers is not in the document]
valeriocos: Yes, that's possible. I ran it from `wg-evolution/bin`

Polaris000:
@valeriocos , I have pushed new changes. The `pdf` and `images` output won't work directly in this patch since the required `push_time_series` method is added in #242 (still a draft). Anyhow, I copied a few of those changes on my local copy of this branch and they work. Please have a look at some of my responses to your review above

bin/analyze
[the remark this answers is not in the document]
valeriocos: cool, thanks!

valeriocos:
Thanks @Polaris000 , I'll try to check it tonight or tomorrow morning at most.

Polaris000:
Thanks for the review! I'll work on the suggested changes as soon as possible!

valeriocos:
Thanks @Polaris000 for addressing my comments.

> You are right, @valeriocos. I had kept them since they improved reabability since they tell you default args. Of course, the default values of parameters can be adding in the function docstring. but I thought this serves the purpose. WDYT?

yes, they improve reabability, however having mutable objects as default parameters can be risky (https://docs.python-guide.org/writing/gotchas/). Furthermore, in this specific case, the entrypoint is the bin/analyze and the argparser takes care of settings default params.

I would suggest either to:
- (i) remove the default params (this requires less work)
```
run_metrics(items, categories, date_range, is_code, conds, period='M')
```
- (ii) reshape them
```
def run_metrics(items, categories=None, date_range=None,
                is_code=None, conds=None, period='M'):

categories = [COMMIT_CATEGORY] if categories is None else categories
is_code = [Naive()] if is_code is None else is_code
date_range = (None, None) if date_range is None else date_range
conds = [] if conds is None else conds
# period can be left as it is since it's not mutable
```

Polaris000:
Thanks for the review @valeriocos .
I have fixed the logging bug.

Yes, `ISSUE_METRICS` is supposed to remain empty, since they have not been defined yet. I plan to do that over the weekend. 
Anyways, I have imported issue metrics as they would be once defined. Of course, the script won't run unless that part is commented out.
The script will work once the issue metrics are defined.

Regarding your point on the default arguments, I'm removing them.
Please have a look at my latest iteration if convenient.

bin/analyze
```diff
@@ -0,0 +1,367 @@
...
+    """
+    Fetches data required for the analysis.
+
+    Depending on the category of metrics to be computed, this function
+    calls the Git (commits) and the GitHub backend (issues, pull_requests).
+
+    :param owner: The owner of a repository
+    :param repository: The name of a git repository
+    :param api_token: GitHub API token
+    :categories: A list of metric categories. A combination of
```
valeriocos: typo: missing `:param`

bin/analyze
```diff
@@ -0,0 +1,367 @@
...
+    parser.add_argument("-d", "--debug",
+                        action='store_true')
+
+    parser.add_argument("-w", "--write-to",
+                        default='results_dir',
+                        help="Path for output results")
+    return parser.parse_args()
+
+
+def run_metrics(items, categories, date_range, is_code, conds, period):
```
valeriocos: missing `:param` for categories

valeriocos: `period` is not used in this function, why?

bin/analyze
```diff
@@ -0,0 +1,367 @@
...
+    :param owner: The owner of a repository
+    :param repository: The name of a git repository
+    :param api_token: GitHub API token
+    :categories: A list of metric categories. A combination of
+        COMMIT_CATEGORY, ISSUE_CATEGORY and PULL_REQUEST_CATEGORY.
+
+    :returns data: A dictionary of fetched items, segregated by category.
+    """
+
+    if api_token is None:
```
valeriocos: minor comment, this if statement could be rewritten in a more pythonic way:
`api_token = [] if api_token is None else api_token`

bin/analyze
```diff
@@ -0,0 +1,367 @@
...
+        'PostfixExclude': PostfixExclude(args.postfixes_to_exclude),
+        'DirExclude': DirExclude(args.dirs_to_exclude)
+    }
+
+    is_code = [is_code_options[option] for option in args.is_code]
+
+    configure_logging(args.debug)
+
+    # fetching data
+    items = fetch_data(owner, repo, [args.api_token], args.categories)
```
valeriocos: An error was thrown here, changing this line to `items = fetch_data(owner, repo, args.api_token, args.categories)` solved the problem

bin/analyze
Polaris000: oops! good catch.

bin/analyze
Polaris000: Sorry, that shouldn't be there.

bin/analyze
Polaris000: I am having [this issue](https://github.com/chaoss/grimoirelab-perceval/issues/546). When I use the brackets, I get the right output.
If I remove the brackets, I get this:
```bash
(wg_imp) Polaris000@thinkpad:~/Documents/GSOC/wg-gmd$ analyze -r Polaris000/practise -o json -cat issue pull_request -t xxxx
[2019-08-17 17:43:46,277] - Fetching data
[2019-08-17 17:43:46,678] - Rate limit not initialized: 401 Client Error: Unauthorized for url: https://api.github.com/rate_limit
[2019-08-17 17:43:46,939] - Rate limit not initialized: 401 Client Error: Unauthorized for url: https://api.github.com/rate_limit
[2019-08-17 17:43:47,179] - Rate limit not initialized: 401 Client Error: Unauthorized for url: https://api.github.com/rate_limit
...
..
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.github.com/rate_limit
```

Polaris000:
Thanks @valeriocos for the review! Regarding the issue metrics, maybe I can remove the issue metric imports and once I am done with implementing them, I can add them. This way, the pull request can be merged without waiting for the metrics to arrive. WDYT?

bin/analyze
valeriocos: yes, you are right, I was using an old version of perceval.
Could you update the requirements.txt to include:
- grimoirelab-perceval (probably >=0.12.18)
- grimoirelab-toolkit (probably >=0.1.9)
- fpdf (probably >=1.7.2)

valeriocos:
Good idea @Polaris000! please update the PR and we will make a last pass

bin/analyze
Polaris000: sure! Thanks.

valeriocos:
please @Polaris000 ping me when I can review it, thanks

Polaris000:
You can have a look. The last travis test is running.

Polaris000:
Is it this error:
```
    result['metric'].plot_time_series(self.period)
  File "/home/icameisaticoded/Documents/GSOC/wg-gmd/implementations/code_df/metric.py", line 86, in plot_time_series
    params = self._get_params()
AttributeError: 'CodeChangesLinesGit' object has no attribute '_get_params'

```

Polaris000:
> For the next PR, we should probably include some doc in the README or have a separated README . Maybe it's a topic to be discussed with @jgbarah and @aswanipranjal.

Are you referring to general documentation or anything in particular?
If its the former, #241, once complete, should solve that issue do what you want. That part has been mentioned by Jesus.

Polaris000:
Thanks for the reviews @valeriocos .

valeriocos:
> If it is, that is because #242 has not been merged yet. 

Yes, it is that error. Thank you for the pointer

> Are you referring to general documentation or anything in particular?
> If its the former, #241, once complete, should solve that issue do what you want. That part has been mentioned by Jesus.

general documentation, #241 looks promising :)

PR #208, opened 2019-07-16
Proposed changes to standardized headings (One Metric Proof of Concept)

sgoggins:
If this standardization works for the **evolution** working group I will issue pull requests for the rest of the metrics to be released. 

Use Cases --> Objectives
Filter + Visualization placed under common heading of **Sample Filter and Visualization**
Test Cases (Examples) --> Examples
References --> Resources

### DIFF
--- a/metrics/Code_Changes.md
+++ b/metrics/Code_Changes.md
@@ -33,7 +33,7 @@ Usual aggregators are:
 
 * Count. Total number of changes during the period.
 
-## Specific description: git
+### Specific description: git
 
 In the cases of git, a code change is defined as the part of
 a git commit that "touches" a source file.
@@ -48,7 +48,7 @@ In a set of repositories, the same commit may be present in more than one
 of them. Therefore, for counting unique changes,
 repeated commits should be counted only once.
 
-### Git parameters
+#### Git parameters
 
 Mandatory:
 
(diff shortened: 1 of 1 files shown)


metrics/Code_Changes.md
```diff
@@ -101,7 +103,7 @@ Some useful visualizations are:
 These could be represented as bar charts, with time running in the X axis.
 Each bar would represent a code changes during a certain period (eg, a month).
 
-## Reference Implementation
+## Sample Implementation
```
jgbarah: I would stick to "Reference", in the sense that if you're implementing, this is not just another implementation, that you can use as a sample, but a the reference implementation, that you should have into account, in the sense that the results of your implementation should be the same than this one. This is the whole idea of having reference implementations.

metrics/Code_Changes.md
```diff
@@ -70,7 +70,7 @@ Mandatory:
     those which do not touch files, and in some cases are not considered as
     reflecting a coding activity.
 
-## Use Cases
+## Objectives
```
jgbarah: I'm not sure about "objectives". But I agree that captures the idea better than "use cases", and causes less confusion. So, if cannot find some other word, that's ok with me. Thanks!

jgbarah:
Thanks for the idea, @sgoggins. I think it improves the template. Please, have a look at my two comments: I only have serious objections to one of the changes.

metrics/Code_Changes.md
germonprez: Thanks @jgbarah. The current template uses Sample. Now, I'm not saying that that's right :) I certainly see your point and moving Sample --> Reference in the other WGs wouldn't be difficult.

Issue #179, opened 2019-06-21
[GSoC] Problem of being forced to use relative imports


Polaris000:
The structure of the `implementations` folder (has `code_df` and `notebooks_df`) makes it necessary to use relative imports for the notebooks to work. The discussion can be found [here](https://github.com/chaoss/wg-evolution/pull/177#discussion_r295176363).

The main purpose of this issue is to function as a reminder to fix this issue later on.

Polaris000:
I am closing this since it has been resolved.

Issue #225, opened 2019-07-23
[GSoC] Proposal to add travis CI feature to reference implementations


Polaris000:
The purpose of this issue is to discuss the idea and implementation for adding travis integration to the repository.

valeriocos:
Thank you @Polaris000 to open this issue, I have just pinged @germonprez to get write access on this repo.

GeorgLink:
I could give @valeriocos access on the grounds that he is a very active and trusted GSoC mentor. 

@sgoggins and @jgbarah, as the maintainers for the Evolution WG, are you okay with Valerio getting access to this repo?

GeorgLink:
Update: @valeriocos already has Write permission on the repo.

sgoggins:
Could we hear a little more about the expected use case?

valeriocos:
Sure @sgoggins , the idea of adding CI support for the work done in GSoC came out https://github.com/Polaris000/GSoC_19_Perceval_Implementations/blob/master/IRC_chats/july_16_week6.txt.

Tests and CI can be useful to be sure that the outcome of the metrics implementations are correct. I guess @jgbarah can give more details about it.

GeorgLink:
I upgraded @valeriocos to admin rights on this repo, to enable him. I still think that there is more to be discussed during the next call and maybe some innovation for all CHAOSS WGs

valeriocos:
Thank you for upgrading my permissions @GeorgLink . I guess we can freeze everything until the next call. This shouldn't be a blocker since @Polaris000 can run the tests locally.
@Polaris000 feel free to join the discussion.

Polaris000:
Thanks @GeorgLink. I agree with @valeriocos that this can be put on hold till the next call and when Jesus will return.

sgoggins:
I think we can implement Travis-CI, but before we do, a group of at least two individuals who are willing to manage and maintain and setup that infrastructure on an ongoing basis should be identified.

jgbarah:
Well, I think we're done with all of this. So, closing it.

PR #226, opened 2019-07-23
Support travis

valeriocos:
This PR provides some changes (reorganize imports, creation of setup.py, etc.) to support the execution of CI tests.


valeriocos:
I don't have permission to activate Travis on this repo (so I cannot check if everything works smoothly). Btw the tests work locally (@Polaris000 checked it on his side too).

valeriocos:
This PR tackles issue https://github.com/chaoss/wg-evolution/issues/225

sgoggins:
This kind of makes sense, but I wonder if the reference implementations are actually going to change that much over time. My expectation is that they will *evolve* 😂 slowly. I am not opposed in any way, but I wonder if we might go such a long time before the reference implementation changes that whoever wrote the tests won't be around to help interpret a failure. Merely a thought.

valeriocos:
Thank you @sgoggins for your comment

Polaris000:
This also resolves issue #179.

jgbarah:
> This kind of makes sense, but I wonder if the reference implementations are actually going to change that much over time. My expectation is that they will _evolve_ joy slowly. I am not opposed in any way, but I wonder if we might go such a long time before the reference implementation changes that whoever wrote the tests won't be around to help interpret a failure. Merely a thought.

I understand your concerns, @sgoggins. But the idea of the tests tend to be that: if in the future they fail, somebody will have to research to know why they fail. Usually, it is better to know they are no longer working (even to decide t drop them), than to ignore that, and have a reference implementation which is not working. More details in the referenced issues.

So, if nobody opposes, I will configure the repo to run those tests.

GeorgLink:
I am okay with tests, but I think about implications they have on the
working group.
Is there any technical debt that the WG collects by adding tests?
Does this raise the requirements for all metrics to include tests to be
considered complete?

jgbarah:
> I am okay with tests, but I think about implications they have on the working group. Is there any technical debt that the WG collects by adding tests? Does this raise the requirements for all metrics to include tests to be considered complete?

This is something that of course we can discuss. Tests are (I think) a part of the reference implementation, written to ensure that the code produces the expected output. Of course we can live without them, but that means that checking that changes to the reference implementations are good will require manual intervention: tests will be helpful mostly on that front.

In case this adds tech debt to the future, or just that they don't help, we could decide to drop them in the future. And I don't think this should be a requirement for metrics. Now, even having a reference implementation is not a requirement. Depending on how we evolve, maybe at some point having a reference implementation becomes a requirement, if we all decide so, and maybe later having tests for the reference implementation too. But not for now, not without a general consensus, and I think nobody is even proposing that.

From another point of view, you can also consider this as a proof of concept, to test the idea of having tests in this repo, and later help us to evaluate if the idea of reference implementations, and tests for them, is good enough to extend to other wgs...

What do you thnk?

GeorgLink:
I agree. Thank you. Let's learn more about tests in reference
implementations.

jgbarah:
After @GeorgLink comment, I'm merging this patch, since nobody seems to oppose, and we need it to move forward. Thanks to all of you!

PR #229, opened 2019-07-30
add Known Implementation of Reviews_Duration in GrimoireLab

GeorgLink:
Adding a description of how to get the Reviews_Duration metric in GrimoireLab.


sgoggins:
This issue with detail is that it creates a tight coupling between the metrics definition and a point in time tool release.

Issue #188, opened 2019-06-22
Candidate Release Comments (Issues Closed)


GeorgLink:
This issue was created to collect comments about the upcoming metrics release.

This thread is for comments about **Issues Closed**

GitHub location: https://github.com/chaoss/wg-evolution/blob/master/metrics/Issues_Closed.md

Release candidate: https://chaoss.community/metric-issues-closed

See all release candidates of metrics are at:
https://chaoss.community/metrics-rc/

**Important Dates**

Release Freeze: June 21st, 2019
Candidate Release: June 24th, 2019
Comments Close: July 24th, 2019
Release Date: August 1st, 2019

klumb:
Dead link (Issues are defined as in Issues.) at line 8.

Not sure if this should go to "Issues New"

if so will need an absolute URL path

jgbarah:
Yes, this should be Issues New.

sgoggins:
I think issues related to source code might be identifiable with a different metric. This varies widely in practice, with platforms like GitHub and GitLab supporting linking more organically... Yet I don't think the consistency of use is widespread. 

So, with regards to this definition from the metric (below), my suggestion is that the default assumption that an issue is NOT linked to source code directly. 

"Criteria for source code. Algorithm. Default: all issues are related to source code.

If we are focused on source code, we need a criteria for deciding whether an issues is related to the source code or not."

sgoggins:
Based on @klumb and @jgbarah comments above mine, I am unsure if I commented in the right place. :)

jgbarah:
> Based on @klumb and @jgbarah comments above mine, I am unsure if I commented in the right place. :)
> 
> My intent is to comment on issues:closed.

I think this is the right place to comment on the Issues Closed metric ;-)

jgbarah:
> [...]
> So, with regards to this definition from the metric (below), my suggestion is that the default assumption that an issue is NOT linked to source code directly. We have not gotten to a level of detail where we define metric attributes (explicitly), but I think some indication of a "source code status" being specified would be appropriate. I think the status would have 3 states:

[...]

PR #230, opened 2019-07-31
Fixing the definition to focus on the "related to code" parameter

sgoggins:
Added the clarifications needed per issue number #188


germonprez:
Looks good to me

PR #232, opened 2019-08-01
[GSoC] Add reference implementation for Issues New

Polaris000:
This pull request adds the reference implementation for the pandas version of Issues New metric. 

As of now, the implementation does not have the `reopen as new` boolean parameter, as mentioned in the [definition of the metric](https://github.com/chaoss/wg-evolution/blob/master/metrics/Issues_New.md). An [issue](https://github.com/chaoss/grimoirelab-perceval/issues/563) has been opened to discuss this.


implementations/tests_df/test_issue_github.py
```diff
@@ -0,0 +1,81 @@
...
+import json
+from pandas.util.testing import assert_frame_equal
+
+import sys
+sys.path.append('..')
+
+from code_df.issue_github import IssueGitHub
+
+
+def read_file(path):
```
jgbarah: Can't we reuse this function? It seems pretty common for all these reference implementations, or am i wrong?

Polaris000: Yes, definitely. But that would require another import. For example, grimoirelab-perceval tests has a similar function, but it is quite frequently repeated in the different backend tests.

jgbarah: What do you mean by "anything else that is imported is tested"?

In any case, I don't mean importing it from other packages, but sharing between the tests.

Polaris000: > What do you mean by "anything else that is imported is tested"?

I meant that, except for standard packages and libraries, the only thing the tests import are the modules in the implementations directory, which are tested. Importing a function to read files would mean that that function would go untested. 

Of course, sharing that function explicitly among tests is alright, probably using a common utils_for_tests.py file...

jgbarah: Well, the function is tested by the tests themselves. If you do think it also needs a test, write it, no problem. But it doesn't make sense to have a function in many tests, which is the same function.

Polaris000: Once all open pull requests regarding tests are approved and merged, I'll send a patch with this change. Thanks for the review!

jgbarah:
Would you mind forcing a new push, to make Travis run the tests?

Polaris000:
@jgbarah please have a look at this after #242. Thanks

PR #235, opened 2019-08-04
[GSoC] Add reference implementation for Issues Active

Polaris000:
This pull request adds the reference implementation for the Issues
Active metric using pandas.

### DIFF
--- a/implementations/code_df/issues_active_github.py
+++ b/implementations/code_df/issues_active_github.py
@@ -1,0 +1,122 @@
+from datetime import datetime
+import pandas as pd
+
+from implementations.code_df.utils import read_json_file, str_to_date
+from implementations.code_df.issue_github import IssueGitHub
+
+
+class IssuesActiveGitHub(IssueGitHub):
+    """
+    Class for Issues Active
+    """
+
+    def _flatten(self, item):
+        """
+        Flatten a raw issue fetched by Perceval into a flat dictionary.
+
+        A list with a single flat directory will be returned.
+        That dictionary will have the elements we need for computing metrics.
+        The list may be empty, if for some reason the commit should not
+        be considered.
+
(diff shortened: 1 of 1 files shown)


Polaris000:
Notebooks and tests will be added soon.

implementations/code_df/issues_active_github.py
```diff
@@ -0,0 +1,122 @@
...
+            'category': "issue",
+            'author': item['data']['user']['login'],
+            'created_date': creation_date,
+            'current_status': item['data']['state'],
+            'updated_date': updation_date
+        }
+
+        return [flat]
+
+    def compute(self):
```
jgbarah: I don't think this works. We're defining "active issues" as those that had some change in state or comment during the period, and this is not capturing that... Have in mind that "updated_date", if I'm not wrong, is giving only the last update to the issue...

I think you need to explore Perceval, to see how this information can be captured, I think it is not captured by default.

If needed, maybe @valeriocos can help here.

Polaris000: @jgbarah, I was going to discuss this in today's meeting. As far as I know, Perceval does not handle events too well. You're right about the `updated_date` meaning. I have opened an issue in Perceval which mentions this, but not specifically. I'll reword that issue or open a new one.

jgbarah: OK, let's discuss today.

jgbarah: I think issues in Perceval are still not solved. @Polaris000, if I'm right, what do you think about closing this pr? When Perceval is ready, we could produce a new patch: it will very likely be different from this one, anyway...

Polaris000: Sure. Closing it for now..

PR #237, opened 2019-08-12
[GSoC] Add non-pandas reference implementation for the Reviews Declined metric

Polaris000:
This pull request adds the non-pandas reference implementation for Reviews Declined.
It adds the python script, the notebook and tests.

### DIFF
--- a/implementations/notebooks/reviews_declined_github.ipynb
+++ b/implementations/notebooks/reviews_declined_github.ipynb
@@ -1,0 +1,171 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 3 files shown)


Polaris000:
Please have a look @jgbarah .

Polaris000:
@valeriocos, I think ipython works differently from PYTHONPATH. Even if I run the jupyter command from the root of the package, my cwd would still be the notebooks folder when I run the notebook. Hence, if I have to use the `implementations.scripts....` absolute import, I think I would have to use the `../..`. WDYT?

valeriocos:
Sorry for the late reply. OK, let's merge it then. Thanks

PR #242, opened 2019-08-15
[GSoC] Add the plot_time_series method to implementation modules

Polaris000:
This pull request adds the plot_time_series (to the root metric class) and the `_get_params` method to each of the metric implementations.


Polaris000:
Please have a look @jgbarah, whenever convenient.

PR #244, opened 2019-10-08
[analyze-script] Add markdown support to analyze script

Polaris000:
This patch adds an option to generate a 
markdown script post analysis of metrics.

### DIFF
--- a/implementations/tests/tests_df/test_open_issue_age_github.py
+++ b/implementations/tests/tests_df/test_open_issue_age_github.py
@@ -41,7 +41,7 @@ class TestOpenIssueAgeGitHub(unittest.TestCase):
         """
 
         open_issue_age = OpenIssueAgeGitHub(self.items)
-        expected_mean = 1132.0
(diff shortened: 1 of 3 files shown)


implementations/tests/tests_df/test_open_issue_age_github.py
```diff
@@ -41,7 +41,7 @@ def test_compute(self):
         """
 
         open_issue_age = OpenIssueAgeGitHub(self.items)
-        expected_mean = 1132.0
+        expected_mean = 1179.0
         mean_age = open_issue_age.compute()
```
Polaris000: This constant change is required fpr the tests of Open Issue Age since age is being computed like so:
```
flat['open_issue_age'] = (datetime.now() - flat['created_date']).days
```
In a following update, I'll make `expected_mean` dynamic, making use of `datetime.now()`.

Polaris000: Fixed in #247

PR #250, opened 2019-10-24
Update Code_Changes.md [WIP]

ccarterlandis:
This patch reorganizes the `Code_Changes.md` metric file to follow the new metric template as we discussed in the call on October 16.

I motion to keep this marked as a draft until #245 is merged first, as it introduces some of the same commits already in that PR.

### DIFF
--- a/focus_areas/templates/question_template.md
+++ b/focus_areas/templates/question_template.md
@@ -1,37 +1,21 @@
-# [Area of Analysis]: [Focal Concern]
+## Description
 
-**Question:** placeholder for a variety of channels (choice?)
+## Objectives
 
+## Implementation
 
-## 1. Description
+### Filters (optional)
 
-_This section provides a rationale for why this question is important to growth maturity and decline._
-_Example for how to use this template: [Documentation](./resources/project_places-documentation.md)_
+### Visualizations (optional)
 
-## 2. Sample Objectives
+### Tools providing the metric (optional)
+- formerly: "known or sample implementations"
+- Metric must be currently deployed/available, in contrast to a tool having the "potential" to provide the metric
+- provide direct link to implementation/documentation, if applicable
 
-- _This section provides a list of reasons a community wants to assess this question_
--
+### Data Collection Strategies (optional)
 
+## Success Metrics (optional)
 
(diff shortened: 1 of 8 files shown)


ccarterlandis:
Right off the bat, there are two things I found myself conflicted on:
1. Are additional subheadings not specified in the template allowed? I opted to include them as I feel it improves readability, but I could also think of at least one counterpoint, namely that if we include/allow too many subheadings then there could be a possibility of the original intention of the template being lost. While I'm not sure this would be likely to happen often, it is something that crossed my mind.

2. Should non-applicable headings stay, with no content underneath them? In this case I opted to remove them if there wasn't any content underneath them, but I was **very** on the fence about this and as I type this comment out I'm starting my to doubt my choice. My original reasoning was that leaving a heading with no content served no real purpose other than to confuse a reader and perhaps make them think it was still a work in progress, or that there was an error. 
On the flip side, if we remove that section, newcomers would not likely know a section existed  unless there was a header for it, and if they have information relevant to that section (maybe they're aware of some external literature) then they might miss out on an opportunity where they could contribute.

Please let me know if these ideas don't make sense or have already been discussed/decided elsewhere.

focus_areas/templates/question_template.md
```diff
@@ -1,37 +1,21 @@
...
-
-- _This section provides detailed instructions for how to execute the above listed quantitative methods and evaluate the results_
--
-
-
-## 5. Resources
-
-- _This section provides a list of resources and references_
+## Resources
+- links to external literature, etc
```
GeorgLink: The focus_areas/templates/question_template.md should be modeled after https://github.com/chaoss/metrics/blob/master/activity-metrics/metric-template.md

Please change the template to match the [CHAOSS default](https://github.com/chaoss/metrics/blob/master/activity-metrics/metric-template.md)

GeorgLink:
I agree with your two conclusions:

1) We want to limit the headings and sub-headings to the ones from the template. I assume some metrics have additional information for "Implementation", which is okay, but maybe not list it as a heading.

2) No. Non-applicable headings should be removed. As you describe, empty sections don't serve any purpose.

PR #254, opened 2019-11-01
Switch Code_Changes_Lines.md to new template

ccarterlandis:
This patch reorganizes Code_Changes_Lines.md to follow the new metric
template. It also changes some internal whitespace in Code_Changes.md
to be both consistent with both itself and with Code_Changes_Lines.md.

### DIFF
--- a/metrics/Code_Changes_Lines.md
+++ b/metrics/Code_Changes_Lines.md
@@ -1,7 +1,10 @@
 # Code Changes Lines
-Question: What is the sum of the number of lines touched (lines added plus lines removed) in all changes to the source code during a certain period?
+
+**Question:** What is the sum of the number of lines touched (lines added plus lines removed) in all changes to the source code during a certain period?
+
 
 ## Description
+
 When introducing changes to the source code, developers touch
 (edit, add, remove) lines of the source code files.
 This metric considers the aggregated number of lines touched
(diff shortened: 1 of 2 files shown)


metrics/Code_Changes_Lines.md
```diff
@@ -1,7 +1,10 @@
 # Code Changes Lines
-Question: What is the sum of the number of lines touched (lines added plus lines removed) in all changes to the source code during a certain period?
+
+**Question:** What is the sum of the number of lines touched (lines added plus lines removed) in all changes to the source code during a certain period?
```
germonprez: Don't bold the question

metrics/Code_Changes_Lines.md
```diff
@@ -88,9 +69,9 @@ and likely, actor merging.
...
 
-## Reference Implementation
- 
-## Known Implementations
+
+### Tools Providing the Metric
+
 * [GrimoireLab](https://chaoss.github.io/grimoirelab) provides this metric out of the box.
   - View an example on the [CHAOSS instance of Bitergia Analytics](https://chaoss.biterg.io/app/kibana#/dashboard/f13af0e0-18e5-11e9-ba47-d5cbef43f8d3).  
   - Download and import a ready-to-go dashboard containing examples for this metric visualization from the [GrimoireLab Sigils panel collection](https://chaoss.github.io/grimoirelab-sigils/chaoss-gmd-cde/lines_of_code_changed/).
```
germonprez: Are these visualizations that could be added to the visualization sub-heading above?

sgoggins: For now the effort is only moving stuff around, not adding things. We can definitely add things as well.

metrics/Code_Changes_Lines.md
sgoggins: It was bolded on the code changes metric. We see its not bolded on the template. We can remove that.

PR #295, opened 2020-01-07
[travis-ci] Remove support for Python 3.4 from travis testing

Polaris000:
This patch updates the travis.yml file and removes
support for Python 3.4. Python 3.5 and 3.6 are 
currently being supported.

This also decreases build time by about 3-4 minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "3.4"
   - "3.5"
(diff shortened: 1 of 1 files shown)


Polaris000:
Can we also setup coveralls along with travis-ci? It would help improve test coverage.

valeriocos:
@Polaris000, coveralls support has just been added!

PR #300, opened 2020-01-09
The classification of commits by tag in Code Changes Lines metric

xavierbol:
In a project, we can have a guideline for the writing of message of
commits. For example, if code changes in the commit are for a bug
fix. The message of this commit can be :
"[FIX] add path into sys.path to include correctly scripts for metrics"

With this "[FIX]" tag, we can classify the commits by tag and have all
commits corresponding to a fix.

Signed-off-by: Xavier Bol <xavier.bol@student.umons.ac.be>


xavierbol:
I think it's good

PR #312, opened 2020-01-17
Implementation: Classify commits by tags

xavierbol:
This pull request adds the implementation about the classification of commits by tags for the Code Changes and Code Changes Lines metrics.
This implementation corresponds to the new filter defined into these pull requests : #299 and #300.


implementations/code_df/code_changes_git.py
```diff
@@ -110,3 +111,12 @@ def __str__(self):
     # time-series on a monthly basis considering only master commits
     print("The number of commits created on a monthly basis is:")
     print(changes.time_series(period='M'))
+
+    # considering commits only that contain either [api] or [backend]
+    # into their message
+    tags = ["[api]", "[backend]"]
+    for tag in tags:
```
Polaris000: I think this should be:
```
# considering only those commits that contain either the [api] or the  
# [backend] tag in their messages

```

implementations/code_df/code_changes_git.py
```diff
@@ -110,3 +111,12 @@ def __str__(self):
...
+        print("Code_Changes, only for the commits that contain {} into their message: {}".format(
+            tag,
+            CodeChangesGit(items, conds=[CommitByTag(tag)]).compute()
```
Polaris000: Change the print statement according to the previous review

implementations/code_df/conditions.py
```diff
@@ -254,3 +254,30 @@ def set_commits(self, commits):
         df = df[~df['merge']]
 
         self.included = set(df['hash'].tolist())
+
+
+class CommitByTag(Commit):
+    """
+    Considerer as a commit contains a specific tag into its message.
+    """
+
```
Polaris000: This should be "Consider as included only commits with a specific tag"

implementations/code_df/conditions.py
```diff
@@ -254,3 +254,30 @@ def set_commits(self, commits):
...
+    def __init__(self, tag):
+        self.tag = tag
+
+    def set_commits(self, commits):
+        """
+        Check if a commit contains the tag in this message.
+
+        Exemple: we want to have the commits when the message begins by [perceval] or [backend]
+
```
Polaris000: I think the `set_commits` docstring can be improved to be like the other sub classes of `Commit`.

implementations/scripts/code_changes_git.py
```diff
@@ -63,3 +64,9 @@ def compute(self):
     changes = CodeChangesGit(items, date_range=(date_since, None),
                              conds=[MasterInclude()])
     print("Code_Changes, only for master:", changes.compute())
+
+    tags = ["[api]", "[backend]"]
+    for tag in tags:
+        print("Code_Changes, only for the commits that contain {} into their message: {}".format(
+            tag,
+            CodeChangesGit(items, conds=[CommitByTag(tag)]).compute()))
```
Polaris000: Please update the print statement as discussed for `code_df` code.

implementations/scripts/conditions.py
```diff
@@ -256,3 +256,26 @@ def set_commits(self, commits):
...
+
+        Exemple: we want to have the commits where the message begins by [perceval] or [backend]
+
+        :param tag: tag to check if the commit contains it in this message
+        :returns: True if the commit contains the tag in its message
+        """
+        items = [dict(commit) for commit in commits]
+
+        items = [item for item in items if self.tag in item['message']]
+        self.included = [item['hash'] for item in items]
```
Polaris000: Maybe the `if self.tag in item['message']` can be written as `if item['message'].startswith(self.tag)` ?

implementations/tests/tests/test_conditions.py
```diff
@@ -271,6 +272,20 @@ class correctly identifies merge commits.
         commit = CommitGit(self.items, conds=[MergeExclude()])
         self.assertEqual(merge_exclude.included, commit.conds[0].included)
 
+    def test_set_commits_commit_by_tag(self):
+        """
+        Test whether the set_commits method of the CommitByTag
+        class correctly identifies the commits contain the tag
+        into their message.
+        """
```
Polaris000: I think this should be "... correctly identifies the commits whose messages start with the [api] tag"

implementations/code_df/conditions.py
```diff
@@ -254,3 +254,30 @@ def set_commits(self, commits):
...
+        :returns: True if the commit contains the tag in its message
+        """
+        self.commits = commits
+        df = self.commits
+
+        todo = set()
+        for _, commit in df.iterrows():
+            if self.tag in commit['message']:
+                todo.add(commit['hash'])
+        self.included = todo
```
Polaris000: Maybe you could modify this approach to make it similar to the other sub -classes of Commit by making use of dataframes? Also, I think it would be cleaner to use `.startswith()` instead of `in` for the conditional check.

implementations/code_df/code_changes_git.py
Polaris000: A suggestion: instead of having to use a for loop and creating multiple objects, maybe you could pass the entire tags list to `CommitByTag` so that it includes those commits which satisfy any one of the tags provided in the list, and modify its implementation in `conditions.py` accordingly.

Let me know if this is not clear, and I'll elaborate.

xavierbol:
@Polaris000 I have made only one pull request because the 2 metrics need the implementation in conditions.py. But if you prefer to divide into 2 pull requests, then I'll do this.
Thanks for your comments, I'll update this on Thursday.

implementations/code_df/code_changes_git.py
xavierbol: It's clear, I'll change this.
Thanks!

Polaris000:
> I have made only one pull request because the 2 metrics need the implementation in conditions.py.

My thought process was that if we had one implementation complete with the CommitByTag functionality, we would be in a position to make a more informed choice on how to proceed with other metrics.

> I'll update this on Thursday.

No problem! Do it whenever convenient.

Polaris000:
@jgbarah @valeriocos please share your views on this if possible!

implementations/code_df/code_changes_git.py
xavierbol: After implementing and testing your suggestion, I think it's a good idea if we want to count the commits whose message starts with one of tags, but if we want to compare the number of commits for each tag, then we don't do this without using the for loop.

Polaris000: You're right when you say that. But the goal of the classes in `conditions.py` is to restrict our analysis. Anytime someone wants to compare the relationship between commits with different tags, they can just not pass the `CommitByTag` condition and then perform the analysis on all commits, segregated by tag using a loop if they prefer.

Later on, if we think a cleaner way to do this is required, we could just add segregating methods for each of the conditions mentioned in `conditions.py`. For now, maybe lets just leave it at that for the sake of consistency.

Polaris000:
Thanks for the changes @xavierbol. The PR looks good IMO. I think the only things left to do are **improving code coverage**, **moving Code_changes_lines to another PR** and **making changes to the Code_Changes_git notebook**. See if you have time for the first two (the notebook can be done later), otherwise let me know.

xavierbol:
@Polaris000 I forgot to say you, why isn't there the notebook for code_changes using list in python (into notebooks folder)?
If you want I can create this notebook according to the notebook using dataframe of pandas library.

Polaris000:
@xavierbol thanks for the changes. 
> why isn't there the notebook for code_changes using list

I think its implementation was delayed due to some structural changes in several metric implementations and then it got lost among all the other implementations.

> If you want I can create this notebook according to the notebook using dataframe of pandas library.

Actually, the point of having two sets of notebooks is that one represents pandas implementations and the other uses just plain python. If you want to send the notebook, follow the analysis done in the CodeChangesGit notebook you just pushed (whatever is computed / whatever charts are produced) but implement it similar to [this](https://github.com/chaoss/wg-evolution/blob/master/implementations/notebooks/code_changes_lines_git.ipynb) (Use whatever structures and constructs are used here).

xavierbol:
Okay, I'll do this notebook

xavierbol:
I finished the updating, I hope it's correct for you. 
Thanks for your reviews and your advice !

Polaris000:
LGTM! @valeriocos: what do you think?

Polaris000:
@ccarterlandis: could you merge this?

sgoggins:
@jgbarah : what are your thoughts on merging this one?

Issue #319, opened 2020-01-23
Merging evolution and common metric (time to response)


klumb:
@ccarterlandis,

These metrics are very similar.

Should we merge them for release or try to differentiate them more?

 If we merge, do we want it here in evolution or should we give it to common?

https://github.com/chaoss/wg-evolution/blob/master/metrics/Issue_Response_Time.md
https://github.com/chaoss/wg-common/blob/master/focus-areas/when/time-to-first-response.md

GeorgLink:
> Hi everyone,
> 
> Thanks Tom for your input. I reflected on this scenario some more and understand now that the Common Metric is a generic version of the more specific Evolution Metric. This could be our first example of using a Common Metric to define more specialized metrics in another WG.
> 
> Maybe we can accept having both metrics but link between them. 
> In the Evolution Metric, we can write something like: This metric is a specific case of <Common Metric>.
> And in the Common Metric, we can write something like: This metric was used in defining <Evolution Metric>
> 
> In the future, there may be n:m relationships between metrics.
> 
> I have changed my mind since the Common WG call and now believe we should keep both metrics with above suggested addition.
> 
> Best,
> Georg



My [comment from the mailing list discussion](https://lists.linuxfoundation.org/pipermail/chaoss/2020-January/001038.html)

ccarterlandis:
+1 from me @GeorgLink! Do you want to open a PR with the change? I am also happy to do it 😄

GeorgLink:
If you have the time to do it, please do. Thanks!

klumb:
@ccarterlandis, the text looks good in the PR. I would move it to the description section though.  @GeorgLink thoughts?

ccarterlandis:
@GeorgLink @klumb I moved the text to the description section and merge my PR, please feel free to revert & re-open this issue if you'd like me to change it again!

klumb:
Looks good to me. Perhaps we should continue this discussion in working group meetings.

GeorgLink:
Thank you! 🙏

PR #321, opened 2020-01-24
Update Issue_Response_Time.md

ccarterlandis:
As per the discussion on #319


klumb:
I think the text is good. Should we move this to the description section?

ccarterlandis:
@klumb I am okay putting it in the description section! I put it in references to start with since I wasn't exactly sure which section it would fit best under. I'll go ahead and make that change

PR #326, opened 2020-02-03
Implementation of Issues in Github without pandas structure

xavierbol:
Before this commit, the implementation for metrics about issues didn't
exist without using the pandas structure. This commit adds the structure
of issues in Github using just plain Python.

After this pull request, other pull requests will come to implement the following metrics :

- Issue New
- Issue Closed
- Issue Age
- Issue Resolution Duration

Signed-off-by: Xavier Bol <xavier.bol@student.umons.ac.be>


Polaris000:
@xavierbol: thanks for the patch. I'll get back to you in a few days.

Polaris000:
@xavierbol: An limitation to keep in mind when implementing Issue metrics with Perceval (especially the `reopen_as_new` parameter) is that Perceval does not fetch event data. About half a year ago, we had a discussion on this. Fetching issue data is very inefficient, and though a patch to address this was created, it was ultimately not merged. Thus, it isn't currently possible to implement issue metrics in their entirety.

The issue implementations which are already on the master branch were pushed before the `events data fetch` patch was discarded.

So lets keep this on hold.

The discussion can be found here:
https://github.com/chaoss/grimoirelab-perceval/issues/563
https://github.com/chaoss/grimoirelab-perceval/pull/565

xavierbol:
@Polaris000: Ok, interesting. Thanks for this information !

Polaris000:
@xavierbol: if you want, you could send the `CommitByTag` patch for `CodeChangesLines`. WDYT?

xavierbol:
@Polaris000: I can but I need the CommitByTag class implemented in the other pull request. I think it's better to send this after the pull request about CodeChanges metric will be merged.

Polaris000:
Ah, my bad. We can carry on once this is merged.

sgoggins:
I think the principle purpose of the code examples in this work group is to illustrate examples of how to apply the definitions using the logic of code. So, limitations of Perceval, for example, is not a significant issue for what we are trying to accomplish.

Polaris000:
@sgoggins, thanks for your input. Since the `reopen_as_new` parameter won't work without Perceval fetching the right data, we can either completely scrap the parameter, making it inconsistent with metric definitions or (and I'm leaning towards this) we could disable the parameter and add a note regarding this issue while keeping the code.

@xavierbol, have a look at my review. Meanwhile, I'll add the note to the README.

implementations/scripts/issue_github.py
```diff
@@ -0,0 +1,125 @@
...
+
+        :param items: list of issues fetched by Perceval into a flat dictionary
+        :param date_range: Tuple, containing a range of date
+        :param reopen_as_new: allow to consider that reopen an issue is a new issue
+        """
+        self.since, self.until = date_range
+        super().__init__(items)
+
+        if reopen_as_new:
+            self.items = self._update_with_reopened_items(self.items)
```
Polaris000: Maybe you could set `reopen_as_new` to `False`, and add a comment based on our discussion on why reopen_as_new doesn't work? WDYT?

The doc string of the function can have the extended note for consistency. Let me know if this is not clear.

xavierbol: The `reopen_as_new` parameter is already set to False by default. Okay, I'll add this comment into the doc string.

xavierbol:
@sgoggins: I understand the principal purpose of the code examples in this working group. My goal in this PR, it's to add a structure of Issue in GitHub similar to the existing implementation using pandas structure. And then, I'll add code examples for the different metrics about Issues.
Because currently, there aren't code examples for issues in plain python and also no code examples for some metrics about issue, for instance, [Issue_Resolution_Duration](https://github.com/chaoss/wg-evolution/blob/master/metrics/Issue_Resolution_Duration.md) metric.

xavierbol:
@Polaris000: Thanks for your review ! I have made the changes, I hope it's good for you.

implementations/scripts/issue_github.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Constructor of this class
+
+        :param items: list of issues fetched by Perceval into a flat dictionary
+        :param date_range: Tuple, containing a range of date
+        :param reopen_as_new: allow to consider that reopen an issue is a new issue.
+            A limitation to keep in mind is that Perceval doesn't fetch event data,
+            and this parameter needs the event data to know if the issue has been
+            reopened. Then, the reopen_as_new parameter couldn't be worked if we
+            doesn't have the event data of issues.
```
Polaris000: Thanks for the patch, @xavierbol! I'm really liking this version. Though the doc-string clearly explains what the `reopen_as_new` represents, I think the wording can be slightly improved. Maybe something like this:
```
"""
:param reopen_as_new: A criteria for deciding whether reopened issues
        are considered as new issues. If True, every time an issue is reopened,
        it is treated as a new issue.

		A limitation to keep in mind is that Perceval
		currently does not retrieve the event data necessary 
		for deciding whether or not an issue has been reopened. Thus, this 
		parameter does not work as expected and shouldn't be set to True.
"""
```

implementations/scripts/issue_github.py
Polaris000: @xavierbol, I wanted to permanently set this to False, though now, I think it's better it remains this way, now that we have added the note in the doc-string.

Polaris000:
LGTM!

sgoggins:
@jgbarah : What are your thoughts on merging this?

Issue #329, opened 2020-03-11
New Contributors Metrics Release Candidate


sgoggins:
This issue was created to collect comments about the rolling release of "New Contributors"

This thread is for comments about New Contributors

GitHub location: https://github.com/chaoss/wg-evolution/blob/master/metrics/New_Contributors.md

GeorgLink:
Read about our [Continuous Metric Contribution process](https://github.com/chaoss/governance/blob/master/community-handbook/metrics-release.md).
Feedback period will end April 8, 2020. If no feedback is provided, we will consider this metric ready for release.

sgoggins:
There are two additional metrics under consideration, "new contributors of issues" and "new contributors of commits" that the March 25, 2020 Evolution Group call suggests adding as filters to this metric. We are also going to add all the filters from the Common Group's Contributor filters to this metric. 

The "Sub class" then becomes "New Contributors" with all the same filters as "Contributors". 

1. Repository authors     
2. Issue authors     
3. Code review participants     
4. Mailing list authors     
5. Event participants     
6. IRC authors     
7. Blog authors     
8. By release cycle     
9. Timeframe of activity in the project     
10. Programming languages of the project     
11. Role or function in project

sgoggins:
We have one small pull request change from Georg, and once he resolves conflicts I believe this metric is ready for release. @klumb @GeorgLink

GeorgLink:
Done

foundjem:
LGTM!

Signed-off-By: Armstrong <foundjem@ieee.org>

ccarterlandis:
Review period has ended, thanks all! Ready for release.

rpaik:
The last sentence under Objectives "Welcome new contributors and thank them for their first contribution" seem incomplete?

germonprez:
Thanks @rpaik 

What about something like: 

An increase or decline in new contributors can be an early indicator of project health. Understanding the behavior and barriers of new community members requires knowing who they are -- helping to welcome new contributors and thank them for their efforts.

klumb:
@sgoggins 
Is this metric ready?

sgoggins:
@klumb : Ready.

ElizabethN:
This metric could use a refresh for readability and clarity. Also, the Description and Objectives sections are essentially identical.

PR #330, opened 2020-03-11
fix list of filters for New_Contributors

GeorgLink:
The filters should be displayed as a list. This PR adds the necessary formatting.


GeorgLink:
relates to #329

GeorgLink:
done

PR #334, opened 2020-04-07
New metric Inactive Contributors

ilam56:
First draft suggestion for new metric that may be useful.

### DIFF
--- a/metrics/Inactive_Contributers.md
+++ b/metrics/Inactive_Contributers.md
@@ -17,7 +17,7 @@ The metric will work by:
 3. if the last contribution date is before the cutoff then add them to the inactivity count of the period they last contibuted in.
 4. create list of inactive contributors
 
-** Aggregators:**
+**Aggregators:**
 - inactive: number of inactive contributors
 
 ### Filters
--- a/metrics/Inactive_Contributers.md
+++ b/metrics/Inactive_Contributers.md
@@ -18,7 +18,7 @@ The metric will work by:
 4. create list of inactive contributers
 
 ### Aggregators:
-inactive: number of inactive contributers
+- inactive: number of inactive contributers
 
 ### Filters
 - Minimum commits required to be considered active
--- a/metrics/Inactive_Contributers.md
+++ b/metrics/Inactive_Contributers.md
@@ -4,10 +4,9 @@ Question: How many Contributers have gone inactive over a specific period of tim
 
 ## Description
 A metric that shows how many contributers have stopped committing over a specific period of time. The period of time required for a contributer to be counted as inactive can be decided by a variable and this metric will display the number of contributers that have been labeled as inactive over a given time frame.
-The first few sentences should match the description in the [metrics list](../activity-metrics-list.md).
 
 ## Objectives
-The objective is to determine how many people are losing intrest in the project. This can be useful in determining intrest in the project. It is important to know if many people are leaving the project and have stopped commiting.
+The objective is to determine how many people are losing intrest in the project. This can be useful in determining intrest in the project. User intrest and user retention is an important metric for determing the health of a project and watching how many users are going inactive can help in finding those values. It is important to know if many people are leaving the project and have stopped commiting.
 
 ## Implementation
 The metric will take in 2 variables a period and a cutoff. The period will be the period over which the number of inactive members will be displayed. For example if period=year then it will display the number of contributers that have gone inactive each year. The cutoff will determine how long it takes for a contributer to be labeled as inactive. If a previous contributer has not made any commits for longer than the cutoff, then they will be added to the inactive count of the period of their last commit. Because of this there will be a bit of a delay between when a user stops commiting and a user shows up as inactive thus this metric will not work for days more recent than the cuttoff.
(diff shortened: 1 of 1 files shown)


metrics/Inactive_Contributers.md
```diff
@@ -0,0 +1,33 @@
...
+## Implementation
+The metric will take in 2 variables a period and a cutoff. The period will be the period over which the number of inactive members will be displayed. For example if period=year then it will display the number of contributers that have gone inactive each year. The cutoff will determine how long it takes for a contributer to be labeled as inactive. If a previous contributer has not made any commits for longer than the cutoff, then they will be added to the inactive count of the period of their last commit. Because of this there will be a bit of a delay between when a user stops commiting and a user shows up as inactive thus this metric will not work for days more recent than the cuttoff.
+
+The metric will work by:
+1. getting a list of all contributers
+2. checking the last commit date
+3. if the last commit date is before the cutoff then add them to the inactivity count of the period they last committed in.
+4. create list of inactive contributers
+
+### Aggregators:
```
sgoggins, line 20: THis should be **bold** instead of a third level heading.

klumb:
Please change language from commits to contributions in this document to reflect all types of contribution inactivity.

metrics/Inactive_Contributers.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Objectives
+The objective is to determine how many people are losing intrest in the project. This can be useful in determining intrest in the project. User intrest and user retention is an important metric for determing the health of a project and watching how many users are going inactive can help in finding those values. It is important to know if many people are leaving the project and have stopped commiting.
+
+## Implementation
+The metric will take in 2 variables a period and a cutoff. The period will be the period over which the number of inactive members will be displayed. For example if period=year then it will display the number of contributers that have gone inactive each year. The cutoff will determine how long it takes for a contributer to be labeled as inactive. If a previous contributer has not made any commits for longer than the cutoff, then they will be added to the inactive count of the period of their last commit. Because of this there will be a bit of a delay between when a user stops commiting and a user shows up as inactive thus this metric will not work for days more recent than the cuttoff.
```
sgoggins, line 12: change "commits" to contributions throughout.

metrics/Inactive_Contributers.md
```diff
@@ -0,0 +1,33 @@
+# Inactive_Contributers
+
+Question: How many Contributers have gone inactive over a specific period of time.
+
+## Description
+A metric that shows how many contributers have stopped committing over a specific period of time. The period of time required for a contributer to be counted as inactive can be decided by a variable and this metric will display the number of contributers that have been labeled as inactive over a given time frame.
+
+## Objectives
+The objective is to determine how many people are losing intrest in the project. This can be useful in determining intrest in the project. User intrest and user retention is an important metric for determing the health of a project and watching how many users are going inactive can help in finding those values. It is important to know if many people are leaving the project and have stopped commiting.
```
sgoggins, line 9: Keep the first sentence, and delete the rest.

metrics/Inactive_Contributers.md
```diff
@@ -0,0 +1,33 @@
+# Inactive_Contributers
```
sgoggins, line 1: "Contributors" is misspelled. Also, the name of the metric has no underscore, though the file does.

metrics/Inactive_Contributers.md
```diff
@@ -0,0 +1,33 @@
...
+
+Question: How many Contributers have gone inactive over a specific period of time.
```
sgoggins, line 3: Question should end with a question mark.

metrics/Inactive_Contributers.md
sgoggins, line 12: Change: 
"If a previous contributer has not made any commits for longer than the cutoff, then they will be added to the inactive count of the period of their last commit. Because of this there will be a bit of a delay between when a user stops commiting and a user shows up as inactive thus this metric will not work for days more recent than the cuttoff."

If a contributor has not made a contribution for a length of time longer than the cutoff, they will be counted as inactive. 

Change "period" to "time period", which is the range of dates we are looking for inactive contributors. Change "cuttoff" to "interval" for the length of time a person will be inactive before we count them as inactive.

metrics/Inactive_Contributers.md
sgoggins: Change the first sentence to "The objective is to determine how many people have stopped contributing actively. This could be useful for community managers to determine if key members are losing interest, or are taking a break.

metrics/Inactive_Contributers.md
ilam56: your definition of period is not what I originally intended. I intended it to be more of a graph period. As in if period is a year it would show how many went inactive in each year. Did you still want me to change it? I hadn't really thought too deeply into it as it was just a rough draft so I don't mind the change in definition.

sgoggins: I see what you are saying. What we really mean is "what is the number of days back" you want look altogether. So, for example, the default "time period" could be "all time", or it could be 3 years, for example. And the "interval" that determines inactivity could be 4 months without any commits or issues.  

I think the original may be a little too complex to understand, and the simpler case we propose will be immediately useful for community managers.

foundjem: @sgoggins Thanks for the clarification. It was a little confusing but its clearer now.

Signed-off-By: Armstrong<foundjem@ieee.org>

PR #343, opened 2020-04-13
Fork Analysis

JosephBurris:
This is a simple measure of the forks made off of a repository. I have changed a couple of things to better-fit the metric outline.
- Included the image for the visualization as a link (so as to not have to create a folder).
- Several changes to the formatting of the document in general.


JosephBurris:
I'm not sure what to do about checks that are not successful. I am fairly new to github still.

PR #359, opened 2020-05-20
Removing folders we no longer use

sgoggins:
Increasing consistency with other working groups.


foundjem:
LGTM!

Signed-off-by: Armstrong <foundjem@ieee.org>

PR #362, opened 2020-07-14
Clarify scope, terminology, and grammar

mbbroberg:
Particularly around the variable names. Consistent naming added in this commit. 

The existing data collection strategy does not offer information and is better represented as a TODO.

I propose merging this PR and following up with a rewrite of the last part with a picture and section similar to [Issue Active](https://github.com/chaoss/wg-evolution/blob/master/metrics/Issues_Active.md).


mbbroberg:
@sgoggins wdyt?

sgoggins (approved):

mbbroberg:
This update appears to have merged already. No idea how this issue is still open (nor do I recall opening it in July -- when was July?!). Closing now.

PR #378, opened 2020-10-19
Add "reactions" as parameter on Issues_Closed

GeorgLink:
Reactions like "thumbs-up" on issues can indicate how much the community cares about issues.

At CLS / ATO 2020, John Coghlan said he'd be interested in understanding how much the community cares about the issues they close.

PR #381, opened 2020-10-21
Add reactions to change requests and issues

GeorgLink:
#378 was the first, this PR adds the new filter to more metrics.

The idea is that we want to understand the impact of our work as it matters to community members. We use *reactions* on an issue or change request as an indicator for how much people care about them. We then use the reactions as a filter or aggregator.

### DIFF
--- a/metrics/Change_Requests.md
+++ b/metrics/Change_Requests.md
@@ -31,6 +31,7 @@ or in some contexts "changesets" in the case of Gerrit.
 
 **Aggregators:**
 * Count. Total number of change requests during the period.
+* Reactions. Number of "thumb-ups" or other reactions on issues.
 
 **Parameters:**
 
--- a/metrics/Change_Requests_Accepted.md
+++ b/metrics/Change_Requests_Accepted.md
@@ -33,6 +33,7 @@ corresponds to a single commit.
 **Aggregators:**
 * Count. Total number of accepted change requests during the period.
 * Ratio. Ratio of accepted change requests over total number of change requests during that period.
+* Reactions. Number of "thumb-ups" or other reactions on issues.
 
 **Parameters:**
 * Period of time. Start and finish date of the period during which accepted change requests are considered. Default: forever.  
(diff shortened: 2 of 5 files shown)


sgoggins:
There is something a bit deeper here than a reaction. Dabbish, for example, refers to emoji as 'signals of attention'. So, "reaction", "reaction emoji", or other reaction may need to emerge as its own metrics.

sgoggins:
This could also apply separately to Twitter. Minnesota did a paper showing how each platform presents a slightly different emoji, and those differences communicate different emotional or reaction content than what is intended on the sending platform.

sgoggins:
Big spikes can alert you to the importance of an issue. 
The majority of the time, its more likely connected to sentiment or the coordinating activities of a project. 
Bottom line: Like all of our metrics, they do not stand alone, out of context. 

Question: Is this more a question of "burstiness"?

sgoggins:
(All comments during May 12, 2021 Evolution WG Meeting)

sgoggins:
The Evolution Working Group discussed this in several meetings and determined that reactions will required a "Goal/Question/Metric" approach to appropriate explain. We have added it to the list of metrics to work on in the metric spreadsheet.

sgoggins:
Upon further review, we would like the language changed to: 

"Reaction: The number of emoji that signal attention to communication or contributions within a project."

And then we will merge this PR. 

We are also developing a separate metric called "Emoji Reactions" that is broader in scope.

sgoggins:
@GeorgLink : If you could also resolve the merge conflicts we will be most grateful. :)

sgoggins:
@GeorgLink : Also note that each of the changes previously proposed uses "issue", even if its a change request related filter. The proposed language change addresses this concern and others.

klumb:
also this should be called emoji reaction to match the metric

germonprez:
@sgoggins what is the status of this PR?

ElizabethN:
I'm going to close this PR without merging - I think this is something that will get covered in the recent discussions around Project Influence and Project Awareness.

ElizabethN:
I should also say that if this needs to be reopened and continued, that's also perfectly valid! But these discussions should go in Common now.

PR #406, opened 2021-04-01
Standardized Repo Structure

yash2002109:
Hello everyone,
I have standardized the repo structure as discussed in #405 and #395. 

I have created a new `focus-areas` directory. The released metrics have been sorted into their focus areas in this directory. I have created an `images` subfolder in each of these focus areas to include the images required for the correct display of these metrics.
I have also included README.md files listing down the files and folders present at each level. The [metrics](https://github.com/chaoss/wg-evolution/tree/master/metrics) and [focus_areas](https://github.com/chaoss/wg-evolution/tree/master/focus_areas) folders have been left unmodified.

I have not yet renamed the files to match the template so that this PR doesn't become too messy. That can be done once these changes have been reviewed.

Issue #405, opened 2021-03-31
Reorganize Repository According to CHAOSS Standard


sgoggins:
https://docs.google.com/spreadsheets/d/1tAGzUiZ9jdORKCnoDQJkOU8tQsZDCZVjcWqXYOSAFmE/edit#gid=1004270137

GeorgLink:
Next: remove `focus_areas` and `metrics` folders from the root of the repo.

yash2002109:
@GeorgLink, Should I submit the required PR for this?

GeorgLink:
Yes, a PR would be great!

yash2002109:
I see many metric files in the `metrics` folder that are not listed in the spreadsheet. I hope it's okay to delete those files..?

klumb:
@yash2002109 

Yes, you can delete them.

yash2002109:
Also, I noticed that there is `standard-readme` badge for the [README](https://github.com/chaoss/wg-evolution/blob/master/README.md) file of this working group. I was wondering if we could incorporate the badge in README files of other working groups too?

klumb:
This still needs to be done

https://github.com/chaoss/governance/blob/master/templates/WG-README.md

PR #407, opened 2021-04-14
Delete inessential metric folders & Update README.md

yash2002109:
In ref to #405 and #406, 
I have done the following changes:

- Deleted the `focus_areas` and `metrics` folder.
- Updated README.md to fix links.


sgoggins:
@GeorgLink : Are you good with these changes?

PR #412, opened 2021-06-19
fixed typos, add more explanations in Issue Resolution Duration for Filters

eyehwan:
fixed typos, change key word `proxy` to `dimension`, add more explanations in `Issue Resolution Duration` for **Filters**

### DIFF
--- a/focus-areas/code-development-activity/branch-lifecycle.md
+++ b/focus-areas/code-development-activity/branch-lifecycle.md
@@ -6,7 +6,7 @@ Question: How do projects manage the lifecycle of their version control branches
 This metric makes the lifecycle of version control branches visible. A branch lifecycle includes acts of branch creation and deletion, as well as persistence of version control branches. When writing code, a development team may create multiple branches, focused around specific features. Subsequently, those branches may be merged into more persistent branches, such as the main branch of a repository. Some branches persist for the life of the repository, while others are deleted after code is merged into a more persistent branch.  By understanding these patterns, we can learn how branch creation, destruction, and merging reflect the development practices of a particular project.  A repository’s typical branch lifecycle and management approach can be used to help identify a project’s repository management style.
 
 ## Objectives
-This metric’s objective is to increase the visibility of a repository’s volume and frequency of branch creation and deletion, as well as persistence of version control branches, in the context of other project characteristics. In turn, this helps potential contributors understand how to initiate and sustain participation in a project. This metric may help  potential contributors understand how to initiate and sustain participation in a project. Questions that can be addressed through this metric include:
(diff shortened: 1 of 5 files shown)


klumb:
I am not sure about changing proxy to dimension. They are not synonyms. 

Changing this term changes the narrative and meaning of the text (possibly in a good way).

Was there a discussion about this somewhere?

eyehwan:
> Was there a discussion about this somewhere?

No, this is the place we start the discussion first. :)

> I am not sure about changing proxy to dimension. They are not synonyms.
Changing this term changes the narrative and meaning of the text (possibly in a good way).

"proxy" from my understanding is more like represented as “act on behalf of”, like [Java Dynamic Proxy Classes](https://docs.oracle.com/javase/8/docs/technotes/guides/reflection/proxy.html) or [Proxy server](https://en.wikipedia.org/wiki/Proxy_server).
But here `Although code changes can be a **proxy** to the coding activity of a project`, makes me confused somewhat. 
If I changed it to **dimension**, it looks more clear that `code change is a dimension of measuring the the coding activity of a project`

I am not a native English speaker. I hope that I understand correctly on the these metrics, which is quite important and nice! 👍

klumb:
No worries @eyehwan. It is a good discussion and edits are most certainly welcome. 

Both terms work here but they mean different things. 

Dimension may be the better word but I would like to get some other opinions on it because it does change the meaning of the text. I am not sure if it changes it in a substantial way, but if it does, we would need to send this metric back to review.

@sgoggins @foundjem -  what do you think?

sgoggins (changes requested):
The Evolution working group thinks this line change to code-changes-lines, code-changes, change-requests-accepted, change-requests, and change-requests-declined should be reverted: 

```
  Although code changes can be a proxy to the coding activity of a project,
  Although code changes can be a dimension to the coding activity of a project,
```

sgoggins:
> The Evolution working group thinks this line change to code-changes-lines, code-changes, change-requests-accepted, change-requests, and change-requests-declined should be reverted:
> 
> ```
>   Although code changes can be a proxy to the coding activity of a project,
>   Although code changes can be a dimension to the coding activity of a project,
> ```

During the July 7, 2021 Evolution working group meeting, we discussed these points at length. "Proxy" has a very specific meaning that clearly represents how people use many Evolution metrics as a "proxy for activity". This word will remain, and we welcome your other changes!! THANK YOU!!!

eyehwan:
@sgoggins  Thanks for the declarations! Then I will push another patch for the rest of fixes.

eyehwan:
I have updated another patch, to revert dimension back to proxy.

PR #448, opened 2022-04-08
Adds Data Ethics Paragraph to Implementation

ElizabethN:
Signed-off-by: Elizabeth Barron <elizabeth@naramore.net>


ElizabethN:
After further consideration, I think this metric could use a refresh. I see some things that could be improved. I'm closing this without merging, and we'll add the data use statement during the refresh.

PR #521, opened 2025-02-26
Add Filter by Label 

geekygirldawn:
Updated this metric to add a Filter option to filter by label. This will be referred to in the new Demonstrating Value Practitioner Guide, but it's also a good filter to have for projects that use labels.


germonprez:
Thanks!

